PR #6, opened 2020-07-10
Xenial build support

calgray:
This change adds a travisCI build against the known cluster library versions:
* gcc 6
* cuda 9.0
* cmake  3.15.1

This version of nvcc can't compile googletest so the .cu test files have been excluded from this build. Cuda kernels can still be tested though via linking to the nvcc compiled objects (example in vector_tests.cc)


src/icrar/leap-accelerate-client/Calibrate.cc
```diff
@@ -77,11 +78,11 @@ namespace icrar
         return MetaData();
     }
 
-    void LeapRemoteCalibration(std::istream& input, std::ostream& output, std::optional<int> overrideStations)
+    void LeapRemoteCalibration(std::istream& input, std::ostream& output, boost::optional<int> overrideStations)
     {
         MetaData metadata = ReadMetaData(input);
 
-        if(overrideStations.has_value())
+        if(overrideStations)
```
calgray, line 85: .is_initialized()

src/icrar/leap-accelerate/utils.h
```diff
@@ -33,14 +33,15 @@
...
 #include <memory>
 #include <vector>
 
 #if (__cplusplus >= 202000L)
+#include <optional>
 #include <filesystem>
 #else
+//#include <boost/optional.hpp>
 //#include <boost/filesystem.hpp>
 #endif
```
calgray, line 46: just use boost and trivial to convert later

src/icrar/leap-accelerate/cuda/math/vector.cu
```diff
@@ -23,6 +23,19 @@
...
 #include "vector.cuh"
 
+// template<typename T, int N>
+// void h_add(const std::array<T, N>& a, const std::array<T, N>& b, std::array<T, N>& c)
+// {
+//     h_add(a.data(), b.data(), c.data(), a.size());
+// }
+
+//template void h_add<int, 1>(const std::array<int, 1>& a, const std::array<int, 1>& b, std::array<int, 1>& c) { h_add(a, b, c); }
+//template void h_add<int, 1000>(const std::array<int, 1000>& a, const std::array<int, 1000>& b, std::array<int, 1000>& c) { h_add(a, b, c); }
```
calgray, line 33: this gets cleaned up in https://github.com/ICRAR/leap-accelerate/pull/7

calgray:
only other change I may need is to setup lapack for cpu side SVD. I'll add it in before supporting cluster builds.

PR #7, opened 2020-07-10
Phase Rotate Math Helpers

calgray:
This PR adds a number of cuda math related operations and tests. This also adds some support for compiling eigen on nvcc.


calgray:
Build is passing later in https://github.com/ICRAR/leap-accelerate/pull/16

PR #13, opened 2020-07-20
MetaData Tests

calgray:
This adds testing for parsing MetaData including PhaseMatrix and Invert Functions


calgray:
Build is passing later in https://github.com/ICRAR/leap-accelerate/pull/16

PR #15, opened 2020-07-20
Inverse Function Tests

calgray:


rtobar:
@calgray you'll need to squash those four last commits -- otherwise the addition of those files will remain in the git repository, and cloning it will always be a relatively expensive operation.

calgray:
Thats a good point, it's annoying how the lfs failed the first time due to single quotes vs double quotes.

I'll squash this and recreate the PR on a new a branch.

PR #16, opened 2020-07-22
PseudoInverse Matrix Tests

calgray:
This changeset adds the psuedoinverse matrix function used in calculating metadata.

Additionally I've added a test dataset for running MetaDataTests.


calgray:
@rtobar It looks like github has some strange LFS quota limitations which makes it essentially unusable for CI even though I'll rarely change them.

I don't think gitlab has the same restriction, however I think I might to avoid migrating just yet.

The two options here are:
* checkout the measurement set normal and grow this repo size by ~200Mb
* create a separate repo for test measurement sets and add it as a submodule

calgray:
The last build issues seems to be a version of eigen3 incompatible with newer cuda versions.

Installing a newer version can be done as described on this repo: https://github.com/PRBonn/refusion

rtobar:
I'll have a look now, green lights look nice :+1:

CMakeLists.txt
```diff
@@ -14,6 +14,12 @@ project(${SOLUTION_NAME} LANGUAGES CXX VERSION ${SOLUTION_NAME_VERSION})
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_BUILD_TYPE Debug)
+set(CMAKE_CXX_FLAGS "-O3 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_DEBUG "-O3 -Wall -Wextra")
```
rtobar, line 18: Adding `-O3` directly into CMAKE_CXX_FLAGS is not really what you want (you can achieve by running `cmake -DCMAKE_BUILD_TYPE=Release ....`).

For the warning flags you could check the compiler ID (e.g., `if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")`) before adding them.

CMakeLists.txt
```diff
@@ -74,9 +80,9 @@ endif()
 
 find_package(GSL)
 find_package(LAPACK)
-find_package(Casacore)
+find_package(Casacore REQUIRED)
```
rtobar, line 83: Note that this won't make a difference because the `FindCasacore.cmake` module included under `cmake` doesn't check this. I included a fix for this in the `non-cuda-builds` branch, so no need to worry about it here.

CMakeLists.txt
```diff
@@ -74,9 +80,9 @@ endif()
...
 find_package(Boost)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3 REQUIRED HINTS /usr/local/include/)
```
rtobar, line 85: I'd assume `/usr/local` is already considered for package searching, so it seems unnecessary. Also `/usr/local/include` should contain only C/C++ header files, so no cmake package should be found there.

calgray:
Other CI build configs that could be considered adding is:
*  Warnings as errors build
* No CUDA build

I'll keep in mind that at some point this may want to allow building on a cluster without cuda (systems can install the cuda library even without a gpu) however it's not a priority for now. In one of my projects I added optional library support via an abstract factory however for a functional library it might be best to use stubs for the cuda linked objects. e.g.

```
cuda/vector.h
--
namespace icrar
{
namespace cuda
{
    void add(size_t n, const double* a, const double* b, double* c);
    void add(size_t n, const float* a, const float* b, float* c);
    void add(size_t n, const int* a, const int* b, int* c);

    void add(const device_vector<double>& a, const device_vector<double>& b, device_vector<double>& c);
    void add(const device_vector<float>& a, const device_vector<float>& b, device_vector<float>& c);
    void add(const device_vector<int>& a, const device_vector<int>& b, device_vector<int>& c);

    ...
}
}

cuda/vector.cc
--
#ifndef CUDA_ENABLED
namespace icrar
{
namespace cuda
{
    void add(size_t n, const double* a, const double* b, double* c) { throw std::runtime_error("cuda is disabled"); }
    void add(size_t n, const float* a, const float* b, float* c) { throw std::runtime_error("cuda is disabled"); }
    void add(size_t n, const int* a, const int* b, int* c) { throw std::runtime_error("cuda is disabled"); }

    void add(const device_vector<double>& a, const device_vector<double>& b, device_vector<double>& c) { throw std::runtime_error("cuda is disabled"); }
    void add(const device_vector<float>& a, const device_vector<float>& b, device_vector<float>& c) { throw std::runtime_error("cuda is disabled"); }
    void add(const device_vector<int>& a, const device_vector<int>& b, device_vector<int>& c) { throw std::runtime_error("cuda is disabled"); }

    ...
}
}
#endif
```

src/icrar/leap-accelerate/MetaData.cc
```diff
@@ -86,32 +86,31 @@ namespace icrar
...
 
         //Start calculations
-
-        casacore::Matrix<double> A;
-        casacore::Array<std::int32_t> I;
         casacore::Matrix<double> A1;
         casacore::Array<std::int32_t> I1;
-        std::tie(A, I) = icrar::cpu::PhaseMatrixFunction(a1, a2, 0);
-        casacore::Matrix<double> Ad = icrar::cpu::RightInvert(A);
+        std::tie(A1, I1) = icrar::cpu::PhaseMatrixFunction(a1, a2, 0);
```
rtobar, line 107: `auto [A1, I1] = icrar::cpu::PhaseMatrixFunction(a1, a2, 0);` should do if you are using 17.

The top-level `CMakeLists.txt` has `set(CMAKE_CXX_STANDARD 17)`, but then the inner `CMakeLists.txt` under `src/icrar/leap-accelerate` has ` target_compile_features(${TARGET_NAME} PUBLIC cxx_std_11)`. Which one is it?

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
```diff
@@ -186,13 +186,13 @@ namespace cpu
         }
 
         Matrix<double> A = Matrix<double>(a1.size() + 1, icrar::ArrayMax(a1) + 1);
-        for(auto v : A)
+        for(double& v : A)
```
rtobar, line 189: Isn't there a `fill`-like operation that could be used instead of an explicit loop?

This could also be `auto &` instead of `double &`, then doesn't need to be changed again if you change the type.

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
```diff
@@ -186,13 +186,13 @@ namespace cpu
...
         {
             v = 0;
         }
 
         Vector<int> I = Vector<int>(a1.size() + 1);
-        for(auto v : I)
+        for(int& v : I)
```
rtobar, line 195: Same as above.

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
calgray: I know there is the std::array::fill() method but I think the only similar thing is the std::fill() function for other collections. Happy to change to `std::fill(A.begin(), A.end(), 0);`

When I reimplement MetaData with Eigen there is a special constructors like Eigen::Matrix::Zero() that I'll swap to.

src/icrar/leap-accelerate/MetaData.cc
calgray: I'm surprised g++ 6 didn't complain about the 17 standard for the other targets. As far as I've checked the cluster runs g++ 6 which should fully support up to c++14 according to this: https://gcc.gnu.org/projects/cxx-status.html#cxx2a

I'll try updating all to C++14 test a basic feature from it such as [[deprecated]]

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
rtobar: There's also an assignment operator: https://casacore.github.io/casacore/classcasacore_1_1Matrix.html#a9c44d638f1a937d71369df6cbdbf2eca, so I think it could also be `A = 0;`. Same for the `Vector` just below.

PR #18, opened 2020-07-28
Add support for building software without CUDA

rtobar:
The following set of changes allow the software to be compiled without CUDA. Most of the infrastructure was already in place, it just needed to be used by the code and the tests.

A new job has been added to the Travis matrix to test that this configuration works correctly. Like the other jobs, it builds the software and runs the tests.

A couple of other minor improvements are included in this set of changes, namely:
 * Adding missing instructions to properly support casacore installations in non-standard locations
 * Removing header inclusions that would otherwise inject CUDA dependencies on non-CUDA-related code


src/icrar/leap-accelerate/tests/math/vector_tests.cc
```diff
@@ -58,119 +83,102 @@ class vector_tests : public testing::Test
...
+    }
+#endif // USE_CUDA
+
+    template<typename backend>
+    void test_device_vector_fibonacci(const int n, const int k)
+    {
+        std::vector<int> a = std::vector<int>(n, 1);
+        std::vector<int> b = std::vector<int>(n, 1);
+        std::vector<int> out = std::vector<int>(n, 0);
+        fibonacci(backend{}, a, b, out, k);
```
calgray, line 160: It's a little odd using empty structs in order to overload a function (fibonacci in this case just an example function for a sequence of operations with gpu memory and how to test it)

If there is no compilation overhead to it I suppose its a valid workaround to using a template argument as a namespace, I'll consider it some more before using it in production functions

rtobar: Yes, it's not great, but given that partial function template specialization doesn't exist this is another way of achieving the same that seems to be popular. There's a way to hide from users those empty structs, but would require an extra layer of code, and since this is test code it doesn't make sense to complicate it even further.

rtobar: @calgray I'm not sure what the outcome of this is -- are you happy to merge this, or you'd rather see some further changes?

calgray:
I know the test classes have a high degree of coupling with cuda and I've added more tests since then and merging this will slow down development a bit further to maintain cuda-less builds passing.

Is there a specific reason why we want to have optional cuda support now? The cuda sdk can be installed on systems even without a graphics card.

rtobar:
> Is there a specific reason why we want to have optional cuda support now? The cuda sdk can be installed on systems even without a graphics card

No, I don't think so, but after a conversation some time ago I thought this was desirable, and I decided to give it a try. If that's not really the case I'm also happy with that :), but then we better reflect that requirement in the `CMakeLists.txt`.

There were still a couple of commits in this PR that were not related to CUDA-less builds, but still useful, so I'll prepare a new PR with only those, plus the additional restriction of making CUDA a hard requirement.

calgray:
The only test I'll have merge conflicts with is the PhaseRotateTests, all the vector/matrix/tensor math tests I'm happy to update with the new format

rtobar:
Maybe it's still probably a good idea to separate the two things anyway (CUDA-less builds from other miscellaneous fixes), so I'll still go ahead with this other PR for getting those fixes in first, then worry about CUDA-less builds on a different one ;)

PR #19, opened 2020-08-03
PhaseRotation Tests

calgray:
Here I've added tests around:
* PhaseRotate function
* Copying metadata between host and device

I've also made a head start on some more cuda classes ready for porting PhaseRotate to cuda.


src/icrar/leap-accelerate/tests/algorithm/PhaseRotateTests.cc
```diff
@@ -97,14 +190,14 @@ namespace icrar
...
-                FAIL() << "Excpected std::invalid_argument";
+                FAIL() << "Expected std::invalid_argument";
             }
         }
     };
 
     TEST_F(PhaseRotateTests, DISABLED_PhaseRotateTestCpu) { PhaseRotateTest(false); }
     TEST_F(PhaseRotateTests, DISABLED_PhaseRotateTestCuda) { PhaseRotateTest(true); }
-    TEST_F(PhaseRotateTests, DISABLED_RotateVisibilitiesTestCpu) { RotateVisibilitiesTest(false); }
+    TEST_F(PhaseRotateTests, RotateVisibilitiesTestCpu) { RotateVisibilitiesTest(false); }
```
calgray, line 200: Currently only this test is enabled but subsequent pull requests are up to enable the others

src/icrar/leap-accelerate/cuda/MetaDataCuda.h
```diff
@@ -33,8 +33,14 @@
 #include <casacore/casa/Arrays/Matrix.h>
 #include <casacore/casa/Arrays/Vector.h>
 
+#ifdef __CUDACC_VER__
+#undef __CUDACC_VER__
+#define __CUDACC_VER__ ((__CUDACC_VER_MAJOR__ * 10000) + (__CUDACC_VER_MINOR__ * 100))
+#endif
```
calgray, line 39: This is an annoying workaround for getting an old eigen version to compile with cuda.

More recently I've started cloning and installing an eigen build newer than the xenial default (3.3.7+) which doesn't require the macro redefinition.

src/icrar/leap-accelerate/MetaData.cc
```diff
@@ -167,16 +175,46 @@ namespace icrar
...
-            metadata.freq_inc_hz,
-            metadata.freq_start_hz + metadata.freq_inc_hz * metadata.channels);
-        for(double& v : metadata.channel_wavelength)
+        for(double& v : channel_wavelength)
         {
             v = speed_of_light / v;
         }
     }
+
+    bool MetaData::operator==(const MetaData& rhs) const
```
rtobar, line 192: My reading is that this is required because `avg_data` is stored inside MetaData, but maybe it would be a better idea to store that separately and leave `MetaData` simpler? The fact that you have averaged *data* inside the *metadata* class also suggests things are getting mixed. No need to address it now though, I just wanted to share my thoughts.

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -57,104 +58,76 @@ namespace icrar
...
-        for(int n = 0; n < a1.size(); n++)
-        {
-            if(a1(IPosition(n)) != a2(IPosition(n)))
+            // Loop over channels
+            for(int channel = 0; channel < metadata.GetConstants().channels; channel++)
             {
-                if((refAnt < 0) || ((refAnt >= 0) & ((a1(IPosition(n)) == refAnt) || (a2(IPosition(n)) == refAnt))))
+                double shiftRad = shiftFactor / metadata.GetConstants().channel_wavelength[channel];
+                double rs = sin(shiftRad);
+                double rc = cos(shiftRad);
```
rtobar, line 111: `std::sin` / `std::cos`. Won't make a difference now, but would if you decided to changed the type of `shiftRad`.

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
```diff
@@ -0,0 +1,192 @@
...
+            uvws.push_back(uvw);
+        }
+    }
+
+    void MetaDataCudaHost::SetDD(const casacore::MVDirection& direction)
+    {
+        m_constants.dlm_ra = direction.get()[0] - m_constants.phase_centre_ra_rad;
+        m_constants.dlm_dec = direction.get()[1] - m_constants.phase_centre_dec_rad;
+
+        dd(0,0) = cos(m_constants.dlm_ra) * cos(m_constants.dlm_dec);
```
rtobar, line 106: `std::cos` / `std::sin` across all invocations, or `using std::cos` / `using std::sin` inside the function body.

src/icrar/leap-accelerate/MetaData.cc
```diff
@@ -143,22 +143,30 @@ namespace icrar
...
+        dd3d(IPosition(I2D,1,0)) = sin(dlm_ra) * cos(dlm_dec);
+        dd3d(IPosition(I2D,1,1)) = cos(dlm_ra);
+        dd3d(IPosition(I2D,1,2)) = sin(dlm_ra) * sin(dlm_dec);
 
-        metadata.dd(IPosition(2,0)) = -sin(metadata.dlm_dec);
-        metadata.dd(IPosition(2,1)) = 0;
-        metadata.dd(IPosition(2,2)) = cos(metadata.dlm_dec);
+        dd3d(IPosition(I2D,2,0)) = -sin(dlm_dec);
+        dd3d(IPosition(I2D,2,1)) = 0;
+        dd3d(IPosition(I2D,2,2)) = cos(dlm_dec);
```
rtobar, line 169: `casacore::Matrix` defines an `T &operator()(size_t i1, size_t i2)`, so you should be able to do a more direct indexing in all these assignments; e.g. `dd3d(0, 0) = cos(dlm_ra) * cos(dlm_dec);`

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
rtobar: Better yet, see the next comment below.

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
```diff
@@ -0,0 +1,192 @@
...
+    {
+        double speed_of_light = 299792458.0;
+        m_constants.channel_wavelength = range(
+            m_constants.freq_start_hz,
+            m_constants.freq_inc_hz,
+            m_constants.freq_start_hz + m_constants.freq_inc_hz * m_constants.channels);
+        for(double& v : m_constants.channel_wavelength)
+        {
+            v = speed_of_light / v;
+        }
```
rtobar, line 129: That looks (I think) exactly like the implementation in `icrar::MetaData`, could that code be shared?

The code of `SetDD` is also the same I think, after removing the usage of `IPosition`-based indexing in `icrar::MetaData::SetDD`, so those could also probably be shared.

src/icrar/leap-accelerate/cuda/MetaDataCuda.h
rtobar: I could see the same workaround in a number of places. Would be good to centralize it into a properly named header, e.g. `eigen_cuda_redefinition_workaround.h` so what it does becomes clear.

Does eigen provide compile-time version checks? It it does, it would be good to constrain the macro redefinition only to eigen versions that need it.

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -57,104 +58,76 @@ namespace icrar
...
-            v = 1;
-        }
+            // For baseline
+            const double pi = boost::math::constants::pi<double>();
+            double shiftFactor = -2 * pi * uvw[baseline].get()[2] - metadata.oldUVW[baseline].get()[2]; // check these are correct
+            shiftFactor = shiftFactor + 2 * pi * (metadata.GetConstants().phase_centre_ra_rad * metadata.oldUVW[baseline].get()[0]);
+            shiftFactor = shiftFactor -2 * pi * (direction.get()[0] * uvw[baseline].get()[0] - direction.get()[1] * uvw[baseline].get()[1]);
 
-        int k = 0;
+            if(baseline % 1000 == 1)
```
rtobar, line 101: Mmm, unusual to use `1` instead of `0`, but equally valid I guess :)

src/icrar/leap-accelerate/MetaData.cc
```diff
@@ -143,22 +143,30 @@ namespace icrar
...
+        if(!dd.is_initialized())
+        {
+            dd.reset(casacore::Matrix<double>(3,3));
+        }
+
+        auto& dd3d = dd.value();
+        dlm_ra = direction.get()[0] - phase_centre_ra_rad;
+        dlm_dec = direction.get()[1] - phase_centre_dec_rad;
+
+        dd3d(IPosition(I2D,0,0)) = cos(dlm_ra) * cos(dlm_dec);
```
rtobar, line 159: `std::sin` / `std::cos` across all invocations, or `using std::sin` / `using std::cos` inside the function body. Won't make a difference now, but could do if the input type changes in the future.

src/icrar/leap-accelerate/MetaData.cc
calgray: I'm also not a huge fan a MetaData class and hope to split the uninitialized data out from it. In  a later PR I move this into a `casalib` namespace to keep as a direct port from the python and introduce new a container (with no optional types) in namespace `cpu`

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
calgray: rs and rc are unused variables from the python code. It looks like they were replaced with the complex number multiply so I'll remove them

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
calgray: I updated these to be methods and they probably could be private as SetWv should really be called in the constructor.

For cuda I might remove this as it may be faster to just compute number from an index rather than copy an extra dynamically sized buffer.

I move speed_of_light into a common folder later on, I'll cherry pick that change into here.

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.h
```diff
@@ -25,10 +25,7 @@
 #include "cuda_runtime.h"
 #include "device_launch_parameters.h"
 
-#ifdef __CUDACC_VER__
-#undef __CUDACC_VER__
-#define __CUDACC_VER__ ((__CUDACC_VER_MAJOR__ * 10000) + (__CUDACC_VER_MINOR__ * 100))
-#endif
+#include <icrar/leap-accelerate/common/eigen_3_3_beta_1_2_support.h>
```
calgray, line 28: macro moved here

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
rtobar: If `SetWV` will change in the CUDA version of the code then that's fine. Regarding `SetDD`, after your changes: does it look similar enough that you could de-duplicate the code?

calgray: Both implementations are unit tested with the same set of inputs and output to intentionally allow the optimized algorithm and class diagram to change freely, in a later PR the interface of the optimized implementation changes from the original. I should probably add documentation to some of the classes now. Basically the difference between MetaData (later becomes casalib::MetaData) and MetaDataHost (later becomes cpu::MetaData) 

casalib::MetaData is a (near) direct port of the working python code primarily for behaviour driven testing of other implementations. Once it's behaviour is the same as the python implementation there is no need to change it.

MetaDataHost/cpu::MetaData is metadata for a c++ cpu optimized implementation of leap-cal using eigen in layout mirroring the cuda implementation (free to completely change from casalib::MetaData)

src/icrar/leap-accelerate/MetaData.cc
```diff
@@ -125,14 +125,14 @@ namespace icrar
...
+    void MetaData::CalcUVW(std::vector<MVuvw>& uvws)
     {
-        metadata.oldUVW = uvws;
+        oldUVW = uvws;
         auto size = uvws.size();
         uvws.clear();
         for(int n = 0; n < size; n++)
         {
-            auto uvw = icrar::Dot(uvws[n], metadata.dd);
+            auto uvw = icrar::Dot(uvws[n], dd.value());
```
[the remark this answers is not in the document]
calgray: throwing an exception above this if dd is not set

src/icrar/leap-accelerate/cuda/MetaDataCuda.cu
rtobar: Ohh, I see, I had failed to grasp that intent the first time around, no problems then :)

rtobar:
Thanks for the further clarifications and changes @calgray, please go ahead at your earliest convenience.

PR #20, opened 2020-08-03
PhaseRotate Cuda with Tensors

calgray:
Here I've split various PhaseRotate implementations into seperate namespaces and an attempt at performing PhaseRotate in a cuda kernel.


CMakeLists.txt
```diff
@@ -14,10 +14,11 @@ project(${SOLUTION_NAME} LANGUAGES CXX VERSION ${SOLUTION_NAME_VERSION})
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_BUILD_TYPE Debug)
-set(CMAKE_CXX_FLAGS_DEBUG "-O1 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_DEBUG "-g -O1 -Wall -Wextra")
```
calgray, line 17: this is needed for gdb

rtobar: I think the cmake's gcc toolchain configuration already includes `-g` in its flags. Have you tried building with `-DCMAKE_BUILD_TYPE=Debug`? That should give you a build with the correct flags already, without the need of adding them yourself.

calgray: I remember explicitly defining this variable to add some optimization and extra errors.

Some of the tests are heavily relying on return value optimization and in one case is the difference between taking 4 seconds vs 20 seconds. I'll look at whether there is a more modern way of adding the flags.

src/icrar/leap-accelerate-client/Calibrate.cc
```diff
@@ -64,7 +64,7 @@ namespace icrar
         MeasurementSet ms = MeasurementSet(ms_filename);
         throw std::runtime_error("not implemented");
 
-        MetaData metadata = MetaData(ms);
+        casalib::MetaData metadata = casalib::MetaData(ms);
```
calgray, line 67: I put the functions using casacore matrices in the `icrar::casalib` namespace since the old casacore version has the `casa -> casacore` macro that often caused problems on xenial if I otherwise named it `icrar::casa`.

CMakeLists.txt
rtobar: Using `-DCMAKE_BUILD_TYPE=Release` should take you to `-O3` land. My point is only to avoid adding specific flags in the `CMakeLists.txt` file and use more of the flexibility given by cmake's options to produce builds at different optimization/quality levels.

src/icrar/leap-accelerate-client/Calibrate.cc
calgray: another possible name for casalib is `ported`

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -67,65 +70,139 @@ namespace cuda
...
+        Constants constants,
+        Eigen::Matrix3d dd,
+        double2 direction,
+        double3* uvw, int uvwLength,
+        double3* oldUVW, int oldUVWLegth,
+        cuDoubleComplex* pavg_data, int avg_dataRows, int avg_dataCols)
+    {
+        using VectorXcucd = Eigen::Matrix<cuDoubleComplex, Eigen::Dynamic, 1>;
+        using MatrixXcucd = Eigen::Matrix<cuDoubleComplex, Eigen::Dynamic, Eigen::Dynamic>;
+        using Tensor3Xcucd = Eigen::Matrix<VectorXcucd, Eigen::Dynamic, Eigen::Dynamic>;
```
calgray, line 95: Replacing this with `Eigen::Matrix<Eigen::Map<VectorXcucd>, Eigen::Dynamic, Eigen::Dynamic>` would almost work but cpu code still needs to allocate memory in a single contiguous block.

In https://github.com/ICRAR/leap-accelerate/pull/21 I refactor MetaData to instead use Eigen::Tensor. It's still possible to access a 2D tensor with the matrix methods using Eigen::Map<Eigen::MatrixXd> so I'm convinced it's the cleanest way to handle dynamic memory allocation.

calgray:
Combining with

PR #21, opened 2020-08-03
PhaseAverage with Cuda

calgray:
This is a running kernel for phase averaging. This uses Eigen::Tensor instead of Eigen::Matrix as it offers advanced slicing methods and can even be mapped back to a matrix using Eigen::Map<Eigen::Matrix> and Eigen::Stride.

Tests are currently naive as I realized I still need to read UVW from the MeasurementSet (or hard code the data).


rtobar:
Will have a look tomorrow when back at the office.

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -67,65 +70,139 @@ namespace cuda
...
-            //metadata['nbaseline']=metadata['stations']*(metadata['stations']-1)/2
-            
-            metadata.SetDD(direction);
-            metadata.SetWv();
-            // Zero a vector for averaging in time and freq
-            metadata.avg_data = Eigen::MatrixXcd(integration.baselines, metadata.GetConstants().num_pols);
-            metadata.init = false;
-        }
-        metadata.CalcUVW(uvw);
+    __global__ void g_RotateVisibilities(
```
calgray, line 84: This is the main cuda function

calgray: In later PRs I update this to the following snippet:
```
__global__ void g_RotateVisibilities(
        cuDoubleComplex* pintegration_data, int integration_data_rows, int integration_data_cols, int integration_data_depth,
        int integration_channels,
        int integration_baselines,
        int polarizations,
        icrar::cpu::Constants constants,
        Eigen::Matrix3d dd,
        double2 direction,
        double3* uvw, int uvwLength,
        double3* oldUVW, int oldUVWLegth,
        cuDoubleComplex* pavg_data, int avg_dataRows, int avg_dataCols)
    {
        using Tensor2Xcucd = Eigen::Tensor<cuDoubleComplex, 2>;
        using Tensor3Xcucd = Eigen::Tensor<cuDoubleComplex, 3>;

        auto integration_data = Eigen::TensorMap<Tensor3Xcucd>(pintegration_data, integration_data_rows, integration_data_cols, integration_data_depth);
        auto avg_data = Eigen::TensorMap<Tensor2Xcucd>(pavg_data, avg_dataRows, avg_dataCols);

        // /// loop over baselines
        for(int baseline = 0; baseline < integration_baselines; ++baseline)
        {
            const double pi = CUDART_PI;
            double shiftFactor = -2 * pi * uvw[baseline].z - oldUVW[baseline].z;
            shiftFactor = shiftFactor + 2 * pi * (constants.phase_centre_ra_rad * oldUVW[baseline].x);
            shiftFactor = shiftFactor - 2 * pi * (direction.x * uvw[baseline].x - direction.y * uvw[baseline].y);

            // loop over channels
            for(int channel = 0; channel < constants.channels; channel++)
            {
                double shiftRad = shiftFactor / constants.GetChannelWavelength(channel);
                cuDoubleComplex exp = cuCexp(make_cuDoubleComplex(0.0, shiftRad));

                for(int polarization = 0; polarization < polarizations; polarization++)
                {
                    integration_data(channel, baseline, polarization) = cuCmul(integration_data(channel, baseline, polarization), exp);
                }

                bool hasNaN = false;
                for(int polarization = 0; polarization < polarizations; polarization++)
                {
                    auto n = integration_data(channel, baseline, polarization);
                    hasNaN |= n.x == NAN || n.y == NAN;
                }

                if(!hasNaN)
                {
                    for(int polarization = 0; polarization < polarizations; ++polarization)
                    {
                        avg_data(baseline, polarization) = cuCadd(avg_data(baseline, polarization), integration_data(channel, baseline, polarization));
                    }
                }
            }
        }
    }
```

src/icrar/leap-accelerate/math/linear_math_helper.h
```diff
@@ -62,6 +64,7 @@ namespace icrar
 
         auto m = Eigen::Matrix<T, R, C>();
 
+        // column major to column major
```
rtobar, line 67: If these copying operations of data are all column-major to column-major you could simply use `std::copy()` instead of the hand-rolled loops. Or is the underlying storage differently organize, and it's column- to row-major?

src/icrar/leap-accelerate/math/linear_math_helper.h
```diff
@@ -96,10 +99,11 @@ namespace icrar
 
         auto m = Eigen::Matrix<T, 3, 3>();
 
+        // column major to column major
```
rtobar, line 102: Same comment/question as above

src/icrar/leap-accelerate/math/linear_math_helper.h
```diff
@@ -114,7 +118,7 @@ namespace icrar
         auto v = Eigen::Matrix<T, Eigen::Dynamic, 1>(value.size());
         for(int i = 0; i < value.size(); ++i)
         {
-            v(i) = value(casacore::IPosition(i));
+            v(i) = value(casacore::IPosition(1, i));
```
rtobar, line 121: You can use `Array::operator[](size_t)` here

src/icrar/leap-accelerate/math/linear_math_helper.h
rtobar: And while I'm here, why is there a special version for 3x3 matrices? The code looks exactly like `template<typename T, int R, int C> ConvertMatrix(const casacore::Matrix<T>& value)` so I'm not sure why this exists.

src/icrar/leap-accelerate/model/MetaData.cc
```diff
@@ -158,17 +166,17 @@ namespace icrar
...
-        dd3d(1,2) = std::sin(dlm_ra) * std::sin(dlm_dec);
+        dd3d(1,0) = sin(dlm_ra) * cos(dlm_dec);
+        dd3d(1,1) = cos(dlm_ra);
+        dd3d(1,2) = sin(dlm_ra) * sin(dlm_dec);
 
-        dd3d(2,0) = -std::sin(dlm_dec);
+        dd3d(2,0) = -sin(dlm_dec);
         dd3d(2,1) = 0;
-        dd3d(2,2) = std::cos(dlm_dec);
+        dd3d(2,2) = cos(dlm_dec);
```
rtobar, line 179: Mmmm... I'm sure we went through this, looks like a step in the opposite direction! :)

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -67,65 +70,139 @@ namespace cuda
...
         {
-            // For baseline
-            const double pi = boost::math::constants::pi<double>();
-            double shiftFactor = -2 * pi * uvw[baseline].get()[2] - metadata.oldUVW[baseline].get()[2]; // check these are correct
-            shiftFactor = shiftFactor + 2 * pi * (metadata.GetConstants().phase_centre_ra_rad * metadata.oldUVW[baseline].get()[0]);
-            shiftFactor = shiftFactor -2 * pi * (direction.get()[0] * uvw[baseline].get()[0] - direction.get()[1] * uvw[baseline].get()[1]);
+            const double pi = CUDART_PI;
+            double shiftFactor = -2 * pi * uvw[baseline].z - oldUVW[baseline].z;
+            shiftFactor = shiftFactor + 2 * pi * (constants.phase_centre_ra_rad * oldUVW[baseline].x);
+            shiftFactor = shiftFactor - 2 * pi * (direction.x * uvw[baseline].x - direction.y * uvw[baseline].y);
```
rtobar, line 107: `shiftFactor += ...` and `shiftFactor -= ...` maybe?

rtobar: And maybe it's cheaper also to factor out `2 * pi`, do all additions/subtractions first, then finally multiply by `2 * pi` once only.

rtobar (commented):
Green in TravisCI and in my computer (code compile and tests run). I left a few comments scattered around, but in general looks good. I'm not an expert on what the phase rotation should be doing, so my comments were limited to what I could say with confidence. Otherwise the rest of the comments are about other changes I could see nor necessarily related to the phase averaging.

As a comment, an improvement to make changes smaller to ease inspection would be to create smaller commits/PRs, otherwise the important bits get lots in between the rest.

I also noted that I had issues compiling against eigen, but I'll open a separate PR or issue with all the details.

src/icrar/leap-accelerate/math/linear_math_helper.h
rtobar: Scratch that -- `Array::operator[]` returns `Array<T>`, not `T`.

But maybe you can use `std::copy` instead of having an explicit loop.

.travis/before_install_bionic.sh
```diff
@@ -25,10 +25,19 @@
...
+#tar -xvjf 3.3.7.tar.bz2
+#cd eigen-eigen-323c052e1731 && mkdir build && cd build
+#cmake ../ && sudo make install
+#cd ../../
+#export EIGEN3_DIR=/usr/local/include/
+
+# Eigen 3.3.9
+git clone https://gitlab.com/libeigen/eigen.git
+cd eigen
+git checkout fb0c6868ad8d43e052c9e027b41b3dfe660bb57d
```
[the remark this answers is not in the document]
calgray: It's strange how most eigen installs have a minor compilation issue when importing tensors that causes an error on the lines of "missing header file "IgnoreAnnoyingWarnings.h" which was why I picked the commit version from the online documentation.

There is probably an ugly workaround to get the other versions to compile but I'll leave that to another day if this project needs to become more portable

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -67,65 +70,139 @@ namespace cuda
...
+                    integration_data(channel, baseline)(i) = cuCmul(integration_data(channel, baseline)(i), exp);
+                }
 
-                data(channel, baseline) = v * std::exp(std::complex<double>(0.0, 1.0) * std::complex<double>(shiftRad, 0.0));
-                if(data(channel, baseline).real() == NAN
-                || data(channel, baseline).imag() == NAN)
+                for(int i = 0; i < integration_data(channel, baseline).cols(); i++)
                 {
-                    metadata.avg_data(1, baseline) += data(channel,baseline);
+                    //if(!integration_data(channel, baseline).hasNaN())
```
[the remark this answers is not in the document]
calgray: the cuDoubleComplex type seems to be lacking a lot of operator overrides. hasNaN() generates this error and I didnt bother fixing it as I needed to swap to tensors:
```
error: no operator "==" matches these operands
            operand types are: const cuDoubleComplex == const cuDoubleComplex
          detected during:
...
instantiation of "__nv_bool Eigen::DenseBase<Derived>::hasNaN() const [with Derived=Eigen::Matrix<cuDoubleComplex, -1, 1, 0, -1, 1>]"
```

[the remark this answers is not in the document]
calgray: basically the kernel compiler does not support std::complex<T> and looking online its recommended to use `cuFloatComplex` and `cuDoubleComplex` whose header looks like this: http://pixel.ecn.purdue.edu:8080/purpl/WSJ/projects/DirectionalStippling/include/cuComplex.h 

It might be possible to write C++ style extension operator overrides for this so I'll give it go in a later branch

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
```diff
@@ -62,117 +56,70 @@ namespace icrar
...
-                double shiftRad = shiftFactor / metadata.channel_wavelength[channel];
-                std::complex<double> v = data(channel, baseline);
+                double shiftRad = shiftFactor / metadata.m_constants.channel_wavelength[channel];
+
+                data(channel, baseline) *= std::exp(std::complex<double>(0.0, 1.0) * std::complex<double>(shiftRad, 0.0));
 
-                data(channel, baseline) = v * std::exp(std::complex<double>(0.0, 1.0) * std::complex<double>(shiftRad, 0.0));
-                if(data(channel, baseline).real() == NAN
-                || data(channel, baseline).imag() == NAN)
+                if(!data(channel, baseline).hasNaN())
```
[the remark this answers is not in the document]
calgray: The port flipped the boolean logic which I later discovered. Python code was this:
```
if ~(np.isnan(data[ch][bl]).any()):
                metadata['avg_data'][bl]+=data[ch][bl]
```

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -67,65 +70,139 @@ namespace cuda
...
 
-            // Loop over channels
-            for(int channel = 0; channel < metadata.GetConstants().channels; channel++)
-            {
-                double shiftRad = shiftFactor / metadata.GetConstants().channel_wavelength[channel];
-                std::complex<double> v = data(channel, baseline);
+                cuDoubleComplex exp = cuCexp(make_cuDoubleComplex(0.0, shiftRad));
+                for(int i = 0; i < integration_data(channel, baseline).cols(); i++)
+                {
+                    integration_data(channel, baseline)(i) = cuCmul(integration_data(channel, baseline)(i), exp);
```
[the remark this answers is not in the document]
calgray: from what I'm aware of the cuda complex types are almost the same as float2 and double2. The `+` operator I think exists and `*` does not. I'll double check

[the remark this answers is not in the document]
calgray: correction: both operators are not defined:
```
error: no operator "*" matches these operands
            operand types are: cuDoubleComplex * cuDoubleComplex
error: no operator "+" matches these operands
            operand types are: cuDoubleComplex + cuDoubleComplex
```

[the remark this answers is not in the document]
rtobar: OK, thanks for the clarification :+1:

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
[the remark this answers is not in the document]
rtobar: OK, so this will be eventually tackled, right? Maybe add a TODO to make it explicit that you mean to add the condition.

src/icrar/leap-accelerate/math/linear_math_helper.h
calgray: I worked out both are column major after playing around with them for a while. Some matrixes from Eigen can be made row major using a template flag and they get marked different when printed out.

I'll give std::copy a go as tests should break now if the behaviour changes

src/icrar/leap-accelerate/model/MetaData.cc
calgray: must have clicked the wrong thing when merging, my bad!

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
calgray: I recall I ported parts of this wrong too. I'll update with the corrections

src/icrar/leap-accelerate/math/linear_math_helper.h
calgray: good spot! will remove

src/icrar/leap-accelerate/math/linear_math_helper.h
calgray: I think the iterator concept gets disabled with a static assert for 2D Eigen::Matrix
```
error: invalid use of void expression
   66 |         std::copy(value.begin(), value.end(), m.begin());
...
/usr/local/include/eigen3/Eigen/src/Core/StlIterators.h:407:3: error: static assertion failed: YOU_TRIED_CALLING_A_VECTOR_METHOD_ON_A_MATRIX
  407 |   EIGEN_STATIC_ASSERT_VECTOR_ONLY(Derived);
```

Another approach though is a c-style memcpy() is likely the fastest

CMakeLists.txt
```diff
@@ -14,8 +14,8 @@ project(${SOLUTION_NAME} LANGUAGES CXX VERSION ${SOLUTION_NAME_VERSION})
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_BUILD_TYPE Debug)
-set(CMAKE_CXX_FLAGS_DEBUG "-O1 -Wall -Wextra")
-set(CMAKE_CXX_FLAGS_RELEASE "-O3 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O1 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall -Wextra")
```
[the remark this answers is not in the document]
calgray: In the last two commits I looks like the debug build timed out without -01. I'll create and document a separate issue for optimization settings to prevent this from blocking things

src/icrar/leap-accelerate/math/Integration.h
```diff
@@ -59,7 +64,23 @@ namespace icrar
 
         bool operator==(const Integration& rhs) const
         {
-            return data.isApprox(rhs.data, 0.001)
+            // There should be a nicer way of doing this, using Eigen::Tensor is one of them
```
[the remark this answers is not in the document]
calgray: I've updated this to be more complete for test verification purposes for this branch. This code gets removed later on when I update to using tensors.

src/icrar/leap-accelerate/math/linear_math_helper.h
calgray: I'll leave a note here to consider using std::copy however the same situation applies where Eigen::Matrix has its iterators disabled

rtobar: It's a pity that `Eigen::Matrix` cannot be traversed in the normal way, which is something I would have imagined would have been possible. I'm completely unfamiliar with the Eigen API though, so I'm sure there are good reasons.

CMakeLists.txt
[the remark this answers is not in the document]
rtobar: What I'd suggest for forcing the `-O1` in Travis is doing it on the `cmake` command line of the Travis job (e.g., `cmake -DCMAKE_CXX_FLAGS="-O1" ...` on `.travis/install.sh`, together with a comment) instead of hardcoding it in the `CMakeLists.txt` file, which will impose the flag on anyone building the software with a `Debug` release type.

src/icrar/leap-accelerate/math/linear_math_helper.h
calgray: I'll try whats on this page:
https://eigen.tuxfamily.org/dox-devel/group__TutorialSTL.html

apparently I need to specify a 1D view of the matrix to do so.

rtobar:
All looking fine now :+1:

calgray:
Thanks for the review! I'll merge the changes up to the next PR

PR #22, opened 2020-08-05
CMake build improvements

rtobar:
These are minor improvements to the cmake functionality. They allow to use casacore installations in non-standard locations, and state the CUDA dependency as a required one, better reflecting the current state of the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,10 +73,8 @@ else (DOXYGEN_FOUND)
   message("Doxygen needs to be installed to generate the doxygen documentation")
 endif (DOXYGEN_FOUND)
 
-find_package(CUDA)
-if(CUDA_FOUND)
-  enable_language(CUDA)
-endif()
+find_package(CUDA REQUIRED)
+enable_language(CUDA)
 
 find_package(GSL)
 find_package(LAPACK)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,11 @@ find_package(Casacore REQUIRED)
 find_package(Boost)
 find_package(Eigen3 REQUIRED)
 
+# Our FindCasacore doesn't understand REQUIRED, so we need to manually check
+if (NOT CASACORE_FOUND)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -73,17 +73,20 @@ else (DOXYGEN_FOUND)
...
 find_package(GSL)
 find_package(LAPACK)
 find_package(Casacore REQUIRED)
 find_package(Boost)
 find_package(Eigen3 REQUIRED)
 
+# Our FindCasacore doesn't understand REQUIRED, so we need to manually check
+if (NOT CASACORE_FOUND)
+    message(FATAL_ERROR "Casacore not found, cannot proceed")
+endif()
```
calgray, line 88: There is a custom findCasacore.cmake in the repo with similar message:
`message (FATAL_ERROR "Could not find CASACORE!")`

It sounds like CASACORE_FIND_REQUIRED isn't being set somewhere. Could this code be moved to that file with a todo or issue to use the REQUIRED argument?

rtobar: Ahhh, good point! Yes, if we hardcode the `CASACORE_FIND_REQUIRED` variable inside `FindCasacore.cmake` then we would see that error message. I'll do that then.

I'll also add the TODO to use `REQUIRED` on the `find_package` call, but the solution would probably be to use a newer `FindCasacore.cmake` file that supports this properly. Sadly there's no "standard" or "recommended" version, simply what different projects have come up with during the years (which as far as I've seen is usually more-or-less copy-pasting, and hence stagnation).

rtobar:
At the end I ended up fixing `FindCasacore.cmake` to properly handle the `REQUIRED` keyword (it was only a difference in variable name casing).

Would you want me to do the merge to the master, or you'd rather do it?

calgray:
Usually I'd always let the creator hit the merge in case there are still minor commits after the review, sanity testing or the target branch needs to change.

Good find on the casing, happy for you to merge after a green build.

PR #27, opened 2020-09-01
CPU end to end test running

calgray:
Updated the cpu optimized leap algorithm to use all eigen types and added a end to end/performance test that currently runs in ~1.8 seconds.

The datatypes used in MetaData and Integration are significant as they now only use contiguous matrices and vectors that are safe to use as staging buffers for copying to cuda.


src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -92,21 +88,27 @@ namespace casalib
                 metadata.GetBaselines(),
                 metadata.num_pols));
 
-            assert(48 == queue.front().data.dimension(0)); //metadata.channels
-            assert(8001 == queue.front().data.dimension(1)); //metadata.baselines
-            assert(4 == queue.front().data.dimension(2)); //metadata.polarizations
+#if _DEBUG
```
rtobar, line 91: Better to use the more standard `#ifndef NDEBUG`. And if the dumping to `std::cout` is removed there's not even a need for this, as `assert` is only enabled if `NDEBUG` is not defined.

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -92,21 +88,27 @@ namespace casalib
...
             input_queues.push_back(queue);
             output_integrations.push_back(std::queue<IntegrationResult>());
             output_calibrations.push_back(std::queue<CalibrationResult>());
         }
 
+#if _DEBUG
+        std::cout << "direction count " << directions.size() << std::endl;
+        std::cout << "input count " << input_queues.size() << std::endl;
+#endif
+
```
rtobar, line 111: Same as above, prefer using the standard `NDEBUG` macro, or simply remove the printing to stdout

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -135,10 +137,11 @@ namespace casalib
 
             if(integration.is_initialized())
             {
+#if _DEBUG
                 std::cout << "rotate visibilities" << std::endl;
                 std::cout << "integration_number:" << integration.get().integration_number << std::endl;
                 std::cout << "direction:" << direction.get() << std::endl;
-
+#endif
```
rtobar, line 144: Same here

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -202,20 +205,20 @@ namespace casalib
...
+#if _DEBUG
+		std::cout << "integration_data(4,0,0):" << integration_data(4, 0, 0) << std::endl;
         std::cout << "DD:" << metadata.dd.get() << std::endl;
-        std::cout << "uvw:" << uvw.size() << std::endl;
-        std::cout << "oldUvw:" << metadata.oldUVW.size() << std::endl;
+        //std::cout << "uvw:" << uvw << std::endl;
+        //std::cout << "oldUvw:" << metadata.oldUVW << std::endl;
         std::cout << "phase_centre_ra_rad:" << metadata.phase_centre_ra_rad << std::endl;
         std::cout << "phase_centre_dec_rad:" << metadata.phase_centre_dec_rad << std::endl;
+#endif
```
rtobar, line 221: And same again

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -288,7 +299,10 @@ namespace casalib
                 }
             }
         }
+          
+#if _DEBUG
         std::cout << "metadata.avg_data.get()(0, 0) : " << metadata.avg_data.get()(0, 0) << std::endl;
+#endif
```
rtobar, line 305: same

src/icrar/leap-accelerate/math/linear_math_helper.cc
```diff
@@ -48,6 +48,19 @@ namespace icrar
...
 
+    std::vector<icrar::MVuvw> ToUVWVector(const Eigen::MatrixXd& value)
+    {
+        auto res = std::vector<icrar::MVuvw>();
+        res.reserve(value.rows());
+
+        for(int row = 0; row < value.rows(); ++row)
+        {
+            auto vector = value(row, Eigen::all);
+            res.push_back(icrar::MVuvw(vector(0), vector(1), vector(2)));
```
rtobar, line 59: Should be possible to do `emplace_back(vector(0), vector(1), vector(2))` to avoid extra object constructions

src/icrar/leap-accelerate/math/linear_math_helper.cc
```diff
@@ -48,6 +48,19 @@ namespace icrar
...
     }
 
+    std::vector<icrar::MVuvw> ToUVWVector(const Eigen::MatrixXd& value)
+    {
+        auto res = std::vector<icrar::MVuvw>();
+        res.reserve(value.rows());
+
+        for(int row = 0; row < value.rows(); ++row)
+        {
+            auto vector = value(row, Eigen::all);
```
rtobar, line 58: Is it possible to receive a reference instead of a copy here? If yes that would be a win.

src/icrar/leap-accelerate/math/linear_math_helper.cc
```diff
@@ -78,4 +91,22 @@ namespace icrar
...
+    {
+        return icrar::MVDirection(value.get()[0], value.get()[1], value.get()[2]);
+    }
+
+    std::vector<icrar::MVDirection> ToDirectionVector(const std::vector<casacore::MVDirection>& value)
+    {
+        // see https://stackoverflow.com/questions/33379145/equivalent-of-python-map-function-using-lambda
+        std::vector<icrar::MVDirection> res(value.size()); //TODO: this populates with 0, O(n), need to reserve and use back_inserter
+        res.reserve(value.size());
+        auto lambda = [](casacore::MVDirection x)
```
rtobar, line 105: You already have `ToDirection` above, there doesn't seem to be a need for this lambda

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
```diff
@@ -55,53 +57,129 @@ namespace icrar
...
-        cuda::MetaData& metadata,
-        const casacore::MVDirection& directions,
-        std::queue<Integration>& input,
-        std::queue<IntegrationResult>& output_integrations,
-        std::queue<CalibrationResult>& output_calibrations)
+        cpu::MetaData& metadata,
+        const icrar::MVDirection& direction,
+        std::vector<cpu::Integration>& input,
+        std::queue<cpu::IntegrationResult>& output_integrations,
+        std::queue<cpu::CalibrationResult>& output_calibrations)
```
rtobar, line 105: You already are in the `icrar::cpu` namespace but there are many `cpu::`-prefixed names still, which seems unnecessary. I highlighted this change because I realized here, but I can see a few more below at least, maybe there are more.

src/icrar/leap-accelerate/math/linear_math_helper.cc
```diff
@@ -58,7 +71,7 @@ namespace icrar
         // see https://stackoverflow.com/questions/33379145/equivalent-of-python-map-function-using-lambda
         std::vector<casacore::MVuvw> res(value.size()); //TODO: this populates with 0, O(n), need to reserve and use back_inserter
         res.reserve(value.size());
-        auto lambda = [&](icrar::MVuvw x)
+        auto lambda = [](icrar::MVuvw x)
         {
             return casacore::MVuvw(x(0), x(1), x(2));
         };
```
rtobar, line 77: A more meaningful name would be nice

src/icrar/leap-accelerate/model/casa/MetaData.cc
```diff
@@ -182,6 +184,30 @@ namespace casalib
...
+        
+        dd3d(1,0) = sin(dlm_ra) * cos(dlm_dec);
+        dd3d(1,1) = cos(dlm_ra);
+        dd3d(1,2) = sin(dlm_ra) * sin(dlm_dec);
+
+        dd3d(2,0) = -sin(dlm_dec);
+        dd3d(2,1) = 0;
+        dd3d(2,2) = cos(dlm_dec);
+    }
+
```
rtobar, line 210: Suspiciously familiar again, I'm sure it's taken care of in a different PR or was a wrong merge.

src/icrar/leap-accelerate/ms/MeasurementSet.h
```diff
@@ -51,10 +53,12 @@ namespace icrar
...
 
+        int m_stations;
+
     public:
-        MeasurementSet(std::string filepath);
-        MeasurementSet(const casacore::MeasurementSet& ms);
-        MeasurementSet(std::istream& stream);
+        MeasurementSet(std::string filepath, boost::optional<int> overrideNStations);
+        MeasurementSet(const casacore::MeasurementSet& ms, boost::optional<int> overrideNStations);
+        MeasurementSet(std::istream& stream, boost::optional<int> overrideNStations);
```
rtobar, line 61: These `optional`s could be defaulted to `boost::none` or similar to make "simple" construction easier to use.

rtobar (commented):
First round of reviews! The tests pass in Travis and in my machine, so that's good. Most of the comments are around macro usage and namespace prefixes, with some exceptions.

src/icrar/leap-accelerate/math/linear_math_helper.cc
calgray: I think this uses a special vector class to store the vector offsets but I can just index the matrix directly instead

src/icrar/leap-accelerate/math/linear_math_helper.cc
calgray: that's a good simplification, I might even be able to to make a template that does both the allocation and transform for any vector<T>

src/icrar/leap-accelerate/algorithm/cpu/PhaseRotate.cc
calgray: This is mostly leftover as some of these classes were/are also in the icrar namespace. Currently I still have a casalib::MetaData class floating around in here which still needs to be upgraded once I verify the behaviour is identical to the casalib port

src/icrar/leap-accelerate/model/casa/MetaData.cc
calgray: Ahh, this I quickly added in as a duplicate of the above code for testing and was meaning to take out. I'll replace it with this:
```
    void MetaData::SetDD(const icrar::MVDirection& direction)
    {
        SetDD(ConvertDirection(direction));
    }
```

src/icrar/leap-accelerate/ms/MeasurementSet.h
calgray: I had this defaulted before but then realized all of my measurement sets need this overriden and shouldn't be overlooked. I'll double check whether the repaired measurement set I received yesterday also needs to be overriden.

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
rtobar: Note that `NDEBUG` should be used with `#ifndef` rather than `#ifdef`. One basically does a double negation and says "if this not a not debug build", which is confusing, but that's the standard way to go.

src/icrar/leap-accelerate/ms/MeasurementSet.h
calgray: updated where I needed to override so don't mind either way now

src/icrar/leap-accelerate/math/linear_math_helper.cc
calgray: in a later pull request I add icrar::vector_map(vector, lambda) to help shrink this file further

rtobar:
All looks good now, approved from my side

Issue #30, opened 2020-09-10
Fix eigen includes


rtobar:
This is the issue I mentioned in one of my comments recently and I thought at least I'd open an issue so we don't forget about this.

The way Eigen headers are being included seems to be incorrect, as they are being included with the `eigen3` prefix (e.g., `#include <eigen3/Eigen/Core`) rather than directly (e.g., `#include <Eigen/Core>`). The latter form is how includes appear in the Eigen examples. And even without considering that, the former building leap-accelerate against arbitrary Eigen installations not possible.

I believe the correct way of doing this is to remove all `eigen3` prefixes from all includes, and to link in cmake against the `Eigen3::Eigen` target, which will include the `eigen_installation/include/eigen3` directory in its include directory search path. There was a comment on the `CMakeLists.txt` file about this not working in xenial, but since a known, very specific Eigen version is being used this shouldn't be a problem.

I'm happy to provide a patch, but would be rather invasive as it will touch all the eigen includes. Maybe it's better to wait until things have stabilized a bit.

calgray:
This is something I'd been meaning to do after finding the current Eigen version and it sounds like CMake should be able find the Eigen3::Eigen target now on all builds. I haven't tried building and installing the Eigen headers on the cluster machine yet so there may be some extra setup to deploy there.

Feel free to create a pull request that removes the fix header and updates the eigen paths.

PR #31, opened 2020-09-10
Fix Eigen includes

rtobar:
There were two main problems with how Eigen was used in the project:
firstly, we were not linking yet to the Eigen3::Eigen target with cmake,
which didn't allow us to use arbitrary Eigen installations. On the other
hand the way Eigen headers were included wouldn't have allowed us anyway
to use these arbitrary Eigen installations anyway, as they included an
"eigen3" prefix that doesn't play well with the include directories
contained in the Eigen3::Eigen target.

This commit fixes the way Eigen headers are included by removing the
eigen3 prefix (`find src/ -type f -exec sed -i 's|eigen3/||' {} +`), and
by linking our library to the Eigen3::Eigen target.

This addresses #30.

### DIFF
--- a/src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.h
+++ b/src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.h
@@ -29,7 +29,7 @@
 #define EIGEN_VECTORIZE_GPU 1
 #define EIGEN_CUDACC 1
 #include <icrar/leap-accelerate/common/eigen_3_3_beta_1_2_support.h>
-#include <eigen3/Eigen/Core>
+#include <Eigen/Core>
 
 #include <casacore/ms/MeasurementSets.h>
 #include <casacore/casa/Arrays/Vector.h>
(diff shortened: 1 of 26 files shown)


src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.h
```diff
@@ -29,7 +29,7 @@
 #define EIGEN_VECTORIZE_GPU 1
 #define EIGEN_CUDACC 1
 #include <icrar/leap-accelerate/common/eigen_3_3_beta_1_2_support.h>
```
calgray, line 31: is this header still required?

rtobar: I don't know, you tell me ;), I was only focusing on changing the inclusion of eigen headers. If you think that's not necessary it could be tackled in a separate PR.

calgray: I recall only adding it to support the xenial eigen version. If the repo now sets up eigen 3.3.90 on all platforms it can be taken out. Maybe just add it as a note on the issues page and I'll do the change when I get my laptop back.

calgray (approved):

PR #33, opened 2020-09-11
Improve warning flag handling

rtobar:
There were two problems with how warning flags were added: first, the
quoting was wrong and things broke if users specified any value for
CMAKE_CXX_FLAGS when invoking cmake. Secondly, and less of an issue
given the platforms we are aiming for, warnings were added regardless of
the compiler in use.

This commit fixes those two issues. An ideal solution would be to
collect these in a variable and add them on a per-target basis via
target_compile_options, but that's a task for another day.

Signed-off-by: Rodrigo Tobar <rtobar@icrar.org>


calgray (commented):
Not sure whether these arguments are also supported by Cray compiler but I like the specificity of compatible compilers at the top of the cmakelists. In one of my cross platform projects I used `add_compile_options` for this however the internet seems to suggest `set_target_properties` as the modern approach so happy to create an issue with some details of that update.

Sample from other project:
```
if(MSVC)
 	add_compile_options(/Wall /WX /O2)
else()
 	add_compile_options(-Wall -Werror -O3)
endif()
```

calgray (approved):

PR #36, opened 2020-09-17
Add direction arguments to CLI using rapidjson

calgray:
This change adds support for directional arguments in the command line via using json string format.

Also added is JsonHelper tests and a CommandLineTest which executes the command using known working arguments.


src/icrar/leap-accelerate-cli/tests/CommandLineTests.cc
```diff
@@ -0,0 +1,79 @@
...
+
+        void TestSingle()
+        {
+            std::string command = m_binDir + "LeapAccelerateCLI";
+            command += " -f " + m_srcDir + "testdata/1197638568-32.ms";
+            command += " -d [[1.0,0.0]]";
+
+            std::cout << command << std::endl;
+            ASSERT_EQ(0, system(command.c_str()));
+        }
```
rtobar, line 74: If you want to verify that a full program works as expected, a simpler alternative could be to declare them directly in cmake via `add_test` in `src/icrar/leap-accelerate-cli/CMakeLists.txt`. For example:

```cmake
add_test("cli-help" LeapAccelerateCLI --help)
add_test("cli-single" LeapAccelerateCLI  -f ${PROJECT_SOURCE_DIR}/testdata/1197638568-32.ms -d [[1.0,0.0]])
```

Tests fail is they exit with non-zero, which is what you want. This would remove the need to pass down `PROJECT_BINARY_DIR` and `PROJECT_SOURCE_DIR` down to the code from cmake, plus making it easier to add more tests in the future.

If you want to have all tests go through your unit testing framework that's fine too, I thought maybe you'd like to consider this alternative as well.

CMakeLists.txt
```diff
@@ -72,14 +72,13 @@ else (DOXYGEN_FOUND)
...
 
-find_package(CUDA REQUIRED)
 enable_language(CUDA)
-
 find_package(GSL)
 find_package(LAPACK)
 find_package(Casacore REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Eigen3 REQUIRED)
+find_package(RapidJSON REQUIRED)
```
rtobar, line 81: You are finding the package but the `RapidJSON::RapidJSON` target is not consequently used (it should be linked by the `LeapAccelerateCLI` target I believe). It probably works in the Travis builds because it all gets installed under `/usr`, but when building against a non-standard installation compilation would fail.

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -51,18 +50,34 @@ namespace icrar
...
+        Arguments() {}
     };
 
+    /**
+     * Validated set of command line arguments
+     */
     class ArgumentsValidated
     {
         InputType m_source;
         boost::optional<std::string> m_filePath;
```
rtobar, line 75: This wasn't part of this PR, but I noticed that these two don't really need to be members

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -150,32 +193,41 @@ int main(int argc, char** argv)
...
-        icrar::casalib::Calibrate(args.GetMeasurementSet(), directions, 16001);
-        break;
-    }
-    case ComputeImplementation::eigen:
+    try
     {
-        std::vector<icrar::MVDirection> directions; //ZenithDirection(ms);
-        icrar::cpu::Calibrate(args.GetMeasurementSet(), directions, 16001);
-        break;
+        ArgumentsValidated args = ArgumentsValidated(std::move(rawArgs));
```
rtobar, line 199: The `std::move` doesn't seem necessary given that the `ArgumentsValidated` constructor takes a const ref, not an rvalue. `std::move` could be removed here, or the `ArgumentsValidated` constructor could be changed to accept an rvalue instead.

rtobar (commented):
Only some minor comments, this time around the RapidJSON integration into the build system, tests for the command-line utility, plus some one or two small details; otherwise really nice work :)

src/icrar/leap-accelerate-cli/tests/CommandLineTests.cc
calgray: I remember considering using googletest here as it would allow in future the possibility to redirect std::cout and std::err and assert the returned result of the tests. Not worth doing as of yet though

src/icrar/leap-accelerate-cli/main.cc
calgray: I think I'll update to an rvalue here as my intention for using std::move is so the reader will know not to use rawArgs once the validated arguments object exists

CMakeLists.txt
calgray: This compiled due to rapidjson being a header only library in a common location. I'll update as it looks like this will fail if rapidjson isn't installed to a common location

src/icrar/leap-accelerate-cli/main.cc
calgray: I haven't tried testing this with multiple file types however my reasoning for expanding on the filepath argument is to allow for things like converting a relative path to an absolute path, handling an ssh path and something like a --verbose flag to print these things to the terminal to help a user diagnose potential issues with the input

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -46,55 +45,58 @@ namespace icrar
...
 
         /**
          * Resources
          */
         std::unique_ptr<MeasurementSet> m_measurementSet;
-        std::ifstream m_fileStream;
         std::istream* m_inputStream = nullptr; // Cached reference to the input stream
 
     public:
         ArgumentsValidated(const Arguments&& args)
```
rtobar, line 89: If you choose to receive by rvalue reference then you probably want to move things out of `args` instead of copying, but this isn't happening: While `args` itself is constructed by rvalue ref, it is marked as `const` and then its members are copied rather than moved (e.g., `m_source(args.sources)` rather than `m_source(std::move(args.source))`)

rtobar:
Go for it! :+1:

PR #41, opened 2020-09-25
Cluster Build

calgray:
For release testing I've done some modifications to allow building leap-accelerate on pleiades. The most notable changes done to allow this are:
* using an eigen3 submodule of a custom commit without installing for cmake
* using a rapidJSON submodule for cmake
* provide reliable set of compilation commands in ReadMe


calgray:
I've checked that the codebase builds on the cluster compiler and libraries with the exception of the following in casa/model/MetaData.cc

image

README.md
```diff
@@ -40,11 +40,25 @@ LEAP-Accelerate includes:
...
+`export CUDA_HOME=/usr/local/cuda`
+
+`export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${CUDA_HOME}/lib64:${CUDA_HOME}/extras/CUPTI/lib64`
+
+`export PATH=$PATH:$CUDA_HOME/bin`
+
+`mkdir -p build && cd build`
+
+`cmake ../../ -DCASACORE_ROOT_DIR=/usr/local/casacore -DCUDA_HOST_COMPILER=g++ -DCUDA_TOOLKIT_ROOT_DIR="${CUDA_HOME}" -DCMAKE_BUILD_TYPE=Release`
+
```
rtobar, line 56: These instructions are missing the rapidjson dependency. I installed one locally under /home/rtobar/rapidjson and added `-DCMAKE_PREFIX_PATH=$HOME/local/rapidjson` to my `cmake` command line to get past the problem.

Here you also point to the casacore installation under `/usr/local/casacore`, which is different from that in the `casacore/3.1.2` module. I used `-DCASACORE_ROOT_DIR=$BLDR_CASACORE_BASE_PATH` instead (which points to the 3.1.2 installation), and that yielded a successful compilation. I also tried *not* loading `casacore/3.1.2` and using `-DCASACORE_ROOT_DIR=/usr/local/casacore`, but that indeed fails to link. This is due to the C++ ABI breaking change introduced in gcc 5.1: `/usr/local/casacore/lib/libcasa_casa.so` (compiled most probably with `/usr/bin/g++`, 4.8.5) contains a `casa::capitalize(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)` symbol, but a compilation with a newer compiler will require a `casa::capitalize(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)` symbol (mind the extra `__cxx11`).

In summary, you can't use the installation under `/usr/local/casacore` because you are using g++ >= 5.1, so maybe you can adjust the instructions to point to the `casacore/3.1.2` module installation as described above.

cmake/FindCasacore.cmake
```diff
@@ -83,7 +83,9 @@ if (NOT CASACORE_FOUND)
     ## Search for the library
     find_library (CASACORE_${_casacore_var}_LIBRARY
       NAMES casa_${_casacore_lib} ${_casacore_lib}
-      HINTS ${CASACORE_ROOT_DIR}
+      HINTS
+      ${CASACORE_ROOT_DIR}
+      "/usr/local/casacore"
```
rtobar, line 88: Not really necessary if you were already pointing to casacore manually. I'd opt for removing this and have users make this pointing more explicitly in the `cmake` invocation. Maybe a section in the README file about the most useful cmake variables could have this one listed.

src/icrar/leap-accelerate/model/casa/MetaData.cc
```diff
@@ -92,7 +93,7 @@ namespace casalib
 
         if(pms->field().nrow() > 0)
         {
-            Vector<MDirection> dir;
+            Vector<MDirection> dir; //TODO: Does not link on cluster
```
rtobar, line 96: It does when compiling against a compatible installation/compiler combination.

rtobar (changes requested):
I tried this out on pleiades just now and managed to get a successful compilation and test run (I excluded the GPU tests though, haven't tried them). I left a few comments regarding the instruction installations, but otherwise the code does compile and run.

README.md
calgray: Ahh, was unaware usr/local/casacore pointed to a different version. After merging with master this now depends on RapidJSON so I'll update that now

CMakeLists.txt
```diff
@@ -73,11 +73,17 @@ else (DOXYGEN_FOUND)
...
+
+# Global Libraries
 find_package(Boost REQUIRED)
+find_package(Casacore REQUIRED)
+find_package(LAPACK)
+find_package(GSL)
+
+# Local Libraries
+set(EIGEN3_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/external/eigen-3.3.90)
 find_package(Eigen3 REQUIRED)
```
rtobar, line 85: These two lines can be turned into a single `add_subdirectory(external/eigen-3.3.90)` instead, which in turn means you don't need to add the `cmake/FindEigen3.cmake` file to the repo. In the previous version you actually had `include_directories(external/eigen-3.3.90)` (and removed linking to `Eigen3::Eigen`) which also worked fine - but linking against a target is better.

The same doesn't work with RapidJSON's repository though because it doesn't define a `RapidJSON::RapidJSON` target (https://github.com/Tencent/rapidjson/issues/1687).

rtobar (approved):
I left a small comment on how Eigen's target is determined now that it's a submodule, but if you're happy with how things are leave it as it, it's a small details and I wouldn't want to block this really any further.

CMakeLists.txt
calgray: First time I tried this I had some strange build errors but working now. The only significant difference between the two is that add_subdirectory also builds the eigen3 tests which ctest picks in the test runner.

The tests take a few minutes and flood the console with warnings which I'm not to fond of doing regularly, I suspect there may be a flag to exclude tests somewhere but maybe this is something better for future if I wanted to change the eigen build flags.

PR #42, opened 2020-09-29
Cuda Calibrate Tests Passing

calgray:
This is a changeset to get CUDA leap code running identical to the single threaded implementation to about 6 decimal places.

One of the optimizations also made here is to load the visiibilities and uvws once into a single integration object in order to perform one cuda call per direction. this may not be possible to do with the too many baselines and channels as the gpu would run out of memory, however there is a task to report the compute and memory footprint for beta release where the batching of baselines can be restored afterwards.

This also adds a bugfix where only the first direction in the casa implementation was being calculated correctly.


src/icrar/leap-accelerate/cuda/device_tensor.h
```diff
@@ -55,7 +56,8 @@ namespace cuda
         , m_sizeDim1(sizeDim1)
         , m_sizeDim2(sizeDim2)
         {
-            size_t byteSize = sizeDim0 * sizeDim1 * sizeDim2 * sizeof(T);
+            size_t byteSize = GetByteSize();
+            std::cout << "tensor3:" << sizeDim0 << "x" << sizeDim1 << "x" << sizeDim2 << std::endl;
```
rtobar, line 60: Probably use logging instead? But I'm not sure what's the context this is executing on (host/device)

src/icrar/leap-accelerate/math/casacore_helper.h
```diff
@@ -91,6 +91,7 @@ namespace icrar
         for(T t : value)
         {
             *result_it = lambda(t);
+            result_it++;
         }
```
rtobar, line 95: This looks like `std::transform`, where you wouldn't need to keep track of `result_it`.

In fact, the first two `MapCollection` overloads look very similar to `vector_map`, and could probably be implemented similarly.

src/icrar/leap-accelerate/math/casacore_helper.h
```diff
@@ -104,6 +105,7 @@ namespace icrar
         for(T t : value)
         {
             *result_it = lambda(t);
+            result_it++;
```
rtobar, line 108: Ditto

src/icrar/leap-accelerate/math/math_conversion.cc
```diff
@@ -95,8 +88,28 @@ namespace icrar
...
+        }
+        else
+        {
+            tmp(0) = 0.0;
+        }
+        tmp(1) = std::asin(xyz(2));
+        return tmp;
+    }
+
+    std::string ToMSString(std::chrono::_V2::system_clock::duration duration)
```
rtobar, line 109: If you go for a `timer` class this could be its associated `operator <<`.

Also mind that the `std::chrono::_V2` namespace is implementation-specific, you probably want to use the `std::chrono` one here.

src/icrar/leap-accelerate/math/math_conversion.h
```diff
@@ -138,4 +141,17 @@ namespace icrar
...
+     * 
+     * @param cartesian 
+     * @return Eigen::Vector2d 
+     */
+    Eigen::Vector2d to_polar(const MVDirection& xyz);
+
+    /**
+     * @brief Convert duration to millisecond string
+     */
+    std::string ToMSString(std::chrono::_V2::system_clock::duration duration);
```
rtobar, line 156: Mind the `std::chrono::_V2` namespace

src/icrar/leap-accelerate/model/cpu/CalibrateResult.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    CalibrateResult ToCalibrateResult(casalib::CalibrateResult& result)
+    {
+        auto output_integrations = std::vector<std::vector<IntegrationResult>>();
+        auto output_calibrations = std::vector<std::vector<CalibrationResult>>();
+
+        for(auto& queues : result.first)
+        {
+            int index = output_integrations.size();
+            output_integrations.push_back(std::vector<IntegrationResult>());
```
rtobar, line 57: `emplace_back`

.travis/install.sh
```diff
@@ -34,10 +34,12 @@ cd ${TRAVIS_BUILD_DIR}
 mkdir build
 cd build
 
-CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_CXX_COMPILER=$COMPILER"
+#Debug Build Configuration
+CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_CXX_COMPILER=${COMPILER}"
 CMAKE_OPTIONS="$CMAKE_OPTIONS -DCUDA_TOOLKIT_ROOT_DIR=${CUDA_HOME}"
 CMAKE_OPTIONS="$CMAKE_OPTIONS -DGSL_ROOT_DIR=${GSL_ROOT_DIR}"
 CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_CXX_FLAGS_DEBUG=${CMAKE_CXX_FLAGS_DEBUG} -O1"
+CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_BUILD_TYPE=Debug"
```
rtobar, line 42: You could consider building with `RelWithDebInfo` instead of `Debug`, which should speed up the tests and avoid stalls, which is what's causing two of the three failures in Travis.

rtobar (changes requested):
Sorry for the delay, I just finished the first set of reviews. Most comments are around how profiling and logging are implemented, but I leave them more like recommendations rather than things that *need* to be addressed.

Travis is failing, so that needs to be fixed obviously. I left some comments related to what I think is happening there. There was also an unusual `std::chrono::_V2` namespace being used that should be fixed.

CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ endif (DOXYGEN_FOUND)
 enable_language(CUDA)
 
 # Global Libraries
-find_package(Boost REQUIRED)
+find_package(Boost REQUIRED COMPONENTS log)
```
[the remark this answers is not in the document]
calgray: I think only components that use linked binaries are available here as there is no component for optional. The other components I intended to use are log_setup and filesystem.

.travis/install.sh
calgray: I mostly wanted to build debug to keep running the assert() methods regularly, I'll double check the optimization level is being set correctly before taking out debug.

src/icrar/leap-accelerate/cuda/device_tensor.h
calgray: I think I put this once I started seeing some memory related errors that I later found to be related to allocating ~8x more memory than required but is now resolved for cuda implementation. I should probably terminate on cuda errors now or put a guard here that checks the available cuda memory and throw if the tensor is going to exceed the available memory.

This class is a C++ object oriented utility for the host to handle unsafe cuda memory buffers as the Cuda api is C style and uses void** with cudaMalloc(). There are other other third party libraries that do this like https://github.com/CNugteren/CLCudaAPI but APIs like these tend to involve extra files for setting up kernels that can be invoked. I suspect in future the best interface to for would be combining SYCL and CUDA that also allows for single source compilation.

src/icrar/leap-accelerate/core/logging.cc
```diff
@@ -0,0 +1,50 @@
...
+
+namespace icrar
+{
+namespace log
+{
+    /**
+     * @brief Initializes
+     * 
+     */
+    void Initialize()
```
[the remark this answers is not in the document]
calgray: removed profiling.h

src/icrar/leap-accelerate/algorithm/casa/PhaseRotate.cc
```diff
@@ -78,30 +79,46 @@ namespace casalib
...
             input_queues.push_back(queue);
             output_integrations.push_back(std::queue<IntegrationResult>());
             output_calibrations.push_back(std::queue<CalibrationResult>());
         }
-
+#ifdef PROFILING
+        auto endTime = std::chrono::high_resolution_clock::now();
+        std::cout << "read time: " << ToMSString(endTime - startTime) << std::endl;
+        startTime = std::chrono::high_resolution_clock::now();
+#endif
```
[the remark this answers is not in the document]
calgray: I've moved all profiling conditionals to a profiliing_timer that hopefully gets removed by aggressive optimization

rtobar:
The still-broken build seems to be caused by the problem [described here](https://github.com/tensorflow/tensorflow/issues/10220#issuecomment-352110064) (ultimately [an issue in gcc 5.5](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=76731)). But on that job you are already using g++-6, you could try using gcc-6 as well as the `CMAKE_C_COMPILER` and as the host compiler used by `nvcc`.

rtobar (approved):
I see that things are now building and tests running in Travis, which was the main blocker -- otherwise I don't think there was any major issue after all the initial review, so please go ahead.

PR #43, opened 2020-10-02
YAN-526 Resolve System Dependencies of leap-accelerate on Test Node

mdolensk:
added deploy subfolder with a build script initially supporting Hyades deployment as well as a SLURM submission script

### DIFF
--- a/deploy/submit.sh
+++ b/deploy/submit.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Markus Dolensky, ICRAR, 2020-10-01
+#
+# Usage: sbatch leapjob.sh [OPTIONS]
+#  Options:
+#  -h,--help                   Print this help message and exit
+#  -s,--stations INT           Override number of stations to use in the measurement set
+#  -f,--filepath TEXT          MeasurementSet file path
+#  -d,--directions TEXT        Direction calibrations
+#  -i,--implementation TEXT    Compute implementation type
+#  -c,--config TEXT            Config filepath
+#
+# examples:
+#  sbatch ./bin/leapjob.sh --help
+#  sbatch ./bin/leapjob.sh -f ~/leap-accelerate/testdata/1197638568-32.ms -s 126 -i cuda -d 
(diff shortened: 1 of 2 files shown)


deploy/submit.sh
```diff
@@ -0,0 +1,53 @@
...
+# Markus Dolensky, ICRAR, 2020-10-01
+#
+# Usage: sbatch leapjob.sh [OPTIONS]
+#  Options:
+#  -h,--help                   Print this help message and exit
+#  -s,--stations INT           Override number of stations to use in the measurement set
+#  -f,--filepath TEXT          MeasurementSet file path
+#  -d,--directions TEXT        Direction calibrations
+#  -i,--implementation TEXT    Compute implementation type
+#  -c,--config TEXT            Config filepath
```
calgray, line 12: This may be prone to becoming outdated

deploy/submit.sh
```diff
@@ -0,0 +1,53 @@
...
+#
+# examples:
+#  sbatch ./bin/leapjob.sh --help
+#  sbatch ./bin/leapjob.sh -f ~/leap-accelerate/testdata/1197638568-32.ms -s 126 -i cuda -d
```
calgray, line 16: this data is in an absolution path, is it possible to make it a relative path?

deploy/submit.sh
dodsonicrar: Does not seem that this should be a blockage. More like tech. debt

dodsonicrar:
approved :)

deploy/submit.sh
mdolensk: @calgray , had missed the comment before the PR was merged: Yes, I thought about just saying that it passes through the args to leap-accelerate, rather than exposing them explicitly and might do that in the next release.

deploy/submit.sh
mdolensk: @calgray, it's a usage example. The value of option -f is passed to the script and that value can be relative or absolute.

PR #52, opened 2020-10-29
Add E2E SKA Tests

calgray:
simulated SKA Data differs from MWA in that autocorrelations are store in the rows. This change now downloads and runs SKA data and ensures it produces a calibration.


rtobar (approved):
It looks all good!

PR #55, opened 2020-10-30
Reduce memory allocations

rtobar:
These are just two commits reducing the amount of memory allocations performed by the `icrar::cpu::MetaData` function. They don't have an effect on the MaxRSS of the application though, which is dominated by the matrix inversion code, but is still a gain.


calgray (approved):
Changes look good, I know something on debug here was unoptimized so hopefully the move operator solved that too.

There is similar code to this in PhaseRotate.cu that does not get used outside of unit tests and I remove in another branch. No need to update there as well.

rtobar:
Thanks @calgray!

PR #57, opened 2020-11-02
Reduce cuda copying

calgray:
Here done some cleanup around object encapsulation and a small improvement to cuda performance by only copying the avg_data buffer from gpu to cpu memory per direction. Benchmarking on my machine this improved:
* single direction SKA wall time 51.41s -> 50.53s
* multi direction SKA wall time 65.45s -> 57.68s


src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -147,9 +147,10 @@ namespace cuda
                 integration.GetIntegrationNumber(),
                 boost::optional<std::vector<casacore::Vector<double>>>());
         }
+
         BOOST_LOG_TRIVIAL(info) << "Copying Metadata from Device";
-        deviceMetadata.ToHost(hostMetadata);
-        
+        deviceMetadata.AvgDataToHost(hostMetadata.avg_data);
```
calgray, line 152: This is the performance enhancement

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -294,15 +295,15 @@ namespace cuda
         );
```
rtobar, line 295: Just above these lines there are a couple of `assert`s that fail to compile with the latest changes, they need to be changed to use `GetAvgData` and `GetVis`

rtobar (changes requested):
I looked at the changes and they look fine. I got a compilation error on my computer, I left a comment on the issue that needs to be solved. Once I get this up and running in hyades I'll run a new set of benchmarks.

rtobar:
I tested the changes now on hyades and indeed the phase rotation time goes down as expected :+1:. I left more comments on YAN-516.

calgray:
I might change one of the CI builds to Debug configuration as I often forget debug compiles extra lines of source

rtobar (approved):

rtobar:
Yeap, good idea, but remember that pure debug builds fail because they take too long to run the test; so maybe just add the `-DNDEBUG` into the `CMAKE_CXX_FLAGS` in the travis `cmake` invocation.

Anyway, I tried the changes in my computer and now it builds, so I've just approved this.

PR #60, opened 2020-11-02
Doxygentest

llwang00:
Please review. This is only for branch of 'doxygentest', will change to 'master' later.


james-strauss-uwa (approved):
Eventually we should move the xml2palette.py code into an ICRAR repo.

calgray (approved):
Is there still a need for supporting html doxygen generation?

llwang00:
> Is there still a need for supporting html doxygen generation?

I don't think so.

rtobar (approved):
Looks fine to me -- with the caveat that Travis builds are not running at the moment given their change in pricing strategy. When/if they run again in the future we'll also need to add the corresponding secure environment variables in the Travis settings for this repository so that the git push command actually works.

llwang00:
> Looks fine to me -- with the caveat that Travis builds are not running at the moment given their change in pricing strategy. When/if they run again in the future we'll also need to add the corresponding secure environment variables in the Travis settings for this repository so that the git push command actually works.

Ok, thanks.

james-strauss-uwa:
Since Travis CI is not running, I have ported this pipeline to a GitHub Action. The new action is in yan-494, so this branch is probably no longer necessary.

Thanks for the work though, the GitHub Action is almost a direct copy of this.

PR #61, opened 2020-11-03
Yan 494

james-strauss-uwa:
This pull request contains three PythonApp components that perform parts of the LEAP DALiuGE workflow.

ProduceConfig

1. Loads a CSV file containing directions
2. Splits the directions into N groups, where N is the multiplicity of the DALiuGE scatter component
3. Produces a JSON config file for each group

CallLeap

1. Loads a JSON config file
2. Builds a command line based on the contents of the config file
3. Executes LeapAccelerateCLI (or if DEBUG==True, calls Sleep for a short random time)
4. Collects the output from LeapAccelerateCLI and writes it to the single output DALiuGE port

LeapGather

1. Loads the JSON from N instances of CallLeap, where N is the multiplicity of the DALiuGE gather component
2. Trivially combines the JSON into a single JSON object
3. Sends the combined JSON to the single output DALiuGE port

Issues:
CallLeap has a hard-coded filename for the measurementSet. This should be read from a parameter, but requires some support in the EAGLE editor before implementing.


leap_drops/CallLeap.py
```diff
@@ -0,0 +1,89 @@
...
+                                    [dlg_batch_input('binary/*', [])],
+                                    [dlg_batch_output('binary/*', [])],
+                                    [dlg_streaming_input('binary/*')])
+
+    # TODO: this measurementSetFilename is not being read by dlg_string_param
+    #       hard-coding it for the moment
+    #measurementSetFilename = dlg_string_param('measurementSetFilename', '')
+    measurementSetFilename = "/Users/james/working/leap-accelerate/testdata/1197638568-32.ms"
+
+    DEBUG = True
```
calgray, line 49: Does this need to be removed?

leap_drops/ProduceConfig.py
```diff
@@ -0,0 +1,129 @@
...
+    numStations = dlg_int_param('number of stations', 1)
+    implementation = dlg_string_param('eigen', '')
+    autoCorrelation = dlg_int_param('auto correlation', 1)
+
+    # should be read from DALiuGE
+    #NUMBER_OF_COPIES = 1
+    #NUM_STATIONS = 126
+    #DIRECTIONS_FILENAME = "directions.csv"
+    #MEASUREMENT_SET_FILENAME = "/Users/james/working/leap-accelerate/testdata/1197638568-32.ms"
+    #IMPLEMENTATION = 'eigen'
```
calgray, line 61: I should mention that I've renamed this implementation to 'cpu' in the command line arguments to make it clearer that it is the preferred implementation. Let me know support for the 'eigen' string is still required

leap_drops/CallLeap.py
```diff
@@ -0,0 +1,89 @@
...
+        if not os.path.isdir(self.measurementSetFilename):
+            raise Exception("Could not find measurement set directory:" + self.measurementSetFilename)
+
+        # read config from input
+        config = self._readConfig(self.inputs[0])
+        #print(config)
+
+        # build command line
+        commandLine = ['LeapAccelerateCLI', '-f', self.measurementSetFilename, '-s', str(config['numStations']), '-d', str(config['directions'])]
+        #print("commandLine:" + str(commandLine))
```
calgray, line 73: some commented prints are in here

calgray (commented):
Is it possible to also create a ReadMe.md in the `leap_drops/`folder that briefly explains how these scripts integrate with DALiuGE?
This could just be a short summary and a link to some docs on the DALiuGE repo

leap_drops/CallLeap.py
james-strauss-uwa: I've removed the commented line (46). But we do need the hard-coded measurementSetFilename in the short-term. We need to fix an issue in EAGLE before this can be done the "right" way.

leap_drops/ProduceConfig.py
james-strauss-uwa: No, I'll update the components to use 'cpu' instead.

leap_drops/CallLeap.py
james-strauss-uwa: I'll remove these, thanks for the feedback.

james-strauss-uwa:
I think the brief Readme file is a good idea. I'll add one. Thanks.

calgray (approved):
Looks good!

leap_drops/CallLeap.py
calgray: I was mostly trying to point out the DEBUG flag. It sounds like there is a task to update this in future shortly so I'll wait until then before looking it over some more

james-strauss-uwa: I think this DEBUG feature is useful for rapid iterations during debugging, so we'd like leave it in for the short term.

Also, we'd like the content in this pull request to make it to master, so that it will be visible to the Travis CI step that automatically produces the EAGLE components.

Would you be happy to approve (and merge) this PR? It would help us make progress on YAN-407. Thanks.

calgray (approved):
Changes look fine to me. I know the current build failure is unrelated to this PR.

It might also be handy at some point to have a continuous integration test with daliuge but maybe leave that for another time.

PR #63, opened 2020-11-03
Cuda Constant Device Buffer

calgray:
I've noticed that for multiple directions cuda copies large constant matrices from the host to device for each direction which is unnecessary as this buffer can be reused between directions.

From my previous SKA run this improves performance for 9 directions:
wall time: 58.43s -> 52.81s


rtobar:
I just finished testing these changes in hyades -- performance looks awesome! Compared to the original version (i.e., before #57) CUDA phase rotation is now about an order of magnitude faster than before. Great work!

Initial: benchmarks-before-improvement

After this change: benchmarks-after-improvement

rtobar (approved):

calgray:
I suspect 9 directions, CPU metadata loading time is high due to allocating space for 9 integrations rather than reusing the same space like the cuda implementation.

Two options here would be:
* Reuse the buffer to improve the single threaded cpu performance
* Update cpu phase rotate to run with parallel cpu threads

PR #64, opened 2020-11-04
Pull git submodules in cmake

calgray:
I've received the question of how to resolve missing dependencies related to git submodules a number of times so I've done some changes to help automate this build step.

CMake will now pull submodules at configure time and provide an error message if this step fails.


CMakeLists.txt
```diff
@@ -33,21 +33,47 @@ add_definitions(-DTEST_DATA_DIR="${PROJECT_SOURCE_DIR}/testdata")
...
-  # git submodule update
-  # This is a cmake module and needs no further input from you
+# Gather External Projects
+find_package(Git)
+if(GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+  # Update submodules as needed, see https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html
+  option(GIT_SUBMODULE "Check submodules during build" ON)
+  if(GIT_SUBMODULE)
+      message(STATUS "Submodule update")
+      execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
```
rtobar, line 45: The `--recursive` flag seems unnecessary at the moment -- it triggers the cloning and checkout of the googletest submodules of *our* submodules (i.e., `external/CLI11/extern/googletest` and `external/rapidjson-1.1.0/thirdparty/gtest`), both of which are not really required for building our project. Having them there doesn't add anything to our compilation, but I'd be inclined to avoid unnecessarily checking them out.

rtobar (approved):
Nice, I tested it locally and it works. Maybe too eagerly actually, I left a comment about that. Otherwise it looks all well.

It's nice to see also that you've constrained the compilation of gtest to only when needed, I thought that was the case already but I actually hadn't tried it out.

CMakeLists.txt
calgray: I suppose this may want to be avoided if storage space is an issue in some enivironment. I don't mind too much either way.

PR #66, opened 2020-11-06
Corrected Stations Number and Baseline Flagging Support

calgray:
Here I've combined correct station number detection with flagging support in https://github.com/ICRAR/leap-accelerate/pull/58

Calibrations for ska_ionosphere_off are now predominantly around +- 0.1 rads which is expected to be close to 0


src/icrar/leap-accelerate/algorithm/cpu/PhaseMatrixFunction.cc
```diff
@@ -0,0 +1,86 @@
...
+        }
+        if(refAnt < 0)
+        {
+            refAnt = 0;
+        }
+
+        A(k, refAnt) = 1;
+        k++;
+        
+        A.conservativeResize(k-1, Eigen::NoChange);
```
calgray, line 80: this should be k

src/icrar/leap-accelerate/ms/MeasurementSet.cc
```diff
@@ -153,6 +155,35 @@ namespace icrar
...
+        // TODO: may want to consider using logical OR over for each channel and polarization.
+        auto epochIndices = casacore::Slice(0, GetNumBaselines(), 1);
+        auto nBaselines = GetNumBaselines();
+        auto flagSlice = casacore::Slicer(
+            casacore::IPosition(3,0,0,0),
+            casacore::IPosition(3,1,1,nBaselines),
+            casacore::IPosition(3,1,1,1));
+        casacore::Vector<bool> fg = m_msmc->flag().getColumn()
+        (flagSlice).reform(casacore::IPosition(1, nBaselines))
+        (epochIndices);
```
rtobar, line 169: If you are only accessing a slice within each cell, and only certain number of rows, you're better off using `getColumnRange` (which will read only what's needed) instead of `getColumn` + subindexing (which will use more memory). Something like this:

```c++
    Eigen::Matrix<bool, -1, 1> MeasurementSet::GetFlaggedBaselines() const
    {
        // TODO: may want to consider using logical OR over for each channel and polarization.
        auto nBaselines = GetNumBaselines();
        auto rowRange = casacore::Slicer(casacore::Slice(0, nBaselines));
        auto flagSlice = casacore::Slicer(
            casacore::IPosition(2, 0, 0),
            casacore::IPosition(2, 1, 1),
            casacore::IPosition(2, 1, 1));
        casacore::Vector<bool> flags = m_msmc->flag().getColumnRange(rowRange, flagSlice);
        return ToVector(flags);
    }
```

There are other places in the code where the same happens (i.e., a full column is read, then indexed to extract a part of its contents), so little by little we should move towards using the more memory-efficient methods.

src/icrar/leap-accelerate/tests/algorithm/PhaseRotateTests.cc
```diff
@@ -318,10 +324,20 @@ namespace icrar
...
             casacore::Vector<std::int32_t> a1 = msmc->antenna1().getColumn()(epochIndices); 
             casacore::Vector<std::int32_t> a2 = msmc->antenna2().getColumn()(epochIndices);
+            
+            auto flagSlice = casacore::Slicer(
+                casacore::IPosition(3,0,0,0),
+                casacore::IPosition(3,1,1,aSize),
+                casacore::IPosition(3,1,1,1));
+            casacore::Vector<bool> fg = msmc->flag().getColumn()
+            (flagSlice).reform(casacore::IPosition(1, aSize))
+            (epochIndices);
```
rtobar, line 340: Similar comment to that made in `MeasurementSet::GetFlaggedBaselines`

src/icrar/leap-accelerate/model/cpu/MetaData.cc
```diff
@@ -118,45 +118,47 @@ namespace cpu
...
         LOG(info) << "avg_data: " << memory_amount(m_avg_data.size() * sizeof(std::complex<double>));
 
+        auto fg = ms.GetFlaggedBaselines();
+
+        auto uvwShape = msmc->uvw().getColumn().shape();
+        auto uvSlice = casacore::Slicer(
+            casacore::IPosition(2,0,0),
+            casacore::IPosition(2,1,uvwShape[1]),
+            casacore::IPosition(2,1,1));
+        casacore::Array<double> uv = msmc->uvw().getColumn()(uvSlice);
```
rtobar, line 128: I posted a similar comment about how you should switch from reading full columns to reading subsets of the data. Also I think `uvwShape[1]` should be the number of rows of the MS, making the first `getColumn` call unnecessary.

Since here you are reading a slice of the cells of the whole column you should be able to use `getColumn(Slicer)` here rather than the more complex `getColumnRange(Slicer, Slicer)` that I suggested for the other use case.

src/icrar/leap-accelerate/ms/MeasurementSet.cc
```diff
@@ -153,6 +155,35 @@ namespace icrar
...
+        unsigned int sum = 0;
+        for(bool b : flaggedBaselines)
+        {
+            if(b)
+            {
+                sum++;
+            }
+        }
+        return sum;
+    }
```
rtobar, line 185: Does `return std::count(flaggedBaselines.begin(), flaggedBaselines.end(), true);` work?

rtobar (commented):
Looks fine overall, I left a few comments about column access patterns and how they can be improved to avoid using much memory.

src/icrar/leap-accelerate/model/cpu/MetaData.cc
calgray: I'm not quite sure how to rearrange this one

calgray: I don't seem to use uv yet until short baseline filtering

rtobar (approved):

PR #67, opened 2020-11-09
Add Config File Parsing

calgray:
Andreas has mentioned that configuration needs to be done via a configuration file to avoid large scripts with string regex to call leap-accelerate-cli. Here I've added three stages of configuration/argument loading in preference (ordered from highest to lowest):
* CLIArgs
* ConfigArgs
* DefaultArgs

From this change I'll add unit tests around the new behaviour before adding output file support.

### DIFF
--- a/src/icrar/leap-accelerate-cli/Arguments.cc
+++ b/src/icrar/leap-accelerate-cli/Arguments.cc
@@ -38,14 +38,13 @@ namespace icrar
     {
         auto args = CLIArguments();
         args.source = InputType::FILENAME;
-        args.filePath = boost::none; // Measurement set filepath
-        args.configFilePath = boost::none; // Config filepath
+        args.filePath = boost::none;
+        args.configFilePath = boost::none;
         args.outFilePath = boost::none;
 
         args.stations = boost::none;
         args.directions = boost::none;
         args.computeImplementation = std::string("cpu");
-        
         args.mwaSupport = false;
         args.readAutocorrelations = true;
         args.verbosity = static_cast<int>(log::DEFAULT_VERBOSITY);
@@ -217,14 +216,14 @@ namespace icrar
         return m_verbosity;
     }
 
-    Arguments ParseConfig(const std::string& configFilepath)
+    Arguments ArgumentsValidated::ParseConfig(const std::string& configFilepath)
     {
         Arguments args;
         ParseConfig(configFilepath, args);
         return args;
     }
 
-    void ParseConfig(const std::string& configFilepath, Arguments& args)
+    void ArgumentsValidated::ParseConfig(const std::string& configFilepath, Arguments& args)
     {
         auto ifs = std::ifstream(configFilepath);
         rapidjson::IStreamWrapper isw(ifs);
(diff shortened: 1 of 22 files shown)


src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -242,7 +96,8 @@ int main(int argc, char** argv)
     icrar::profiling::UsageReporter _;
     try
     {
-        ArgumentsValidated args = ArgumentsValidated(std::move(rawArgs));
+        ArgumentsValidated args = ArgumentsValidated(std::move(Arguments(std::move(rawArgs))));
```
rtobar, line 99: `ArgumentsValidated args {Arguments{std::move(rawArgs)}};`, no need for the second `move` (but you need list-initialization syntax to avoid the most vexing parse).

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -219,12 +72,13 @@ int main(int argc, char** argv)
     app.set_version_flag("--version", [&]() { return version_information(appName); });
 
     //Parse Arguments
-    Arguments rawArgs;
+    CLIArguments rawArgs = CLIArguments();
```
rtobar, line 75: `CLIArguments rawArgs;`

src/icrar/leap-accelerate-cli/Arguments.cc
```diff
@@ -0,0 +1,359 @@
...
+
+    ArgumentsValidated::ArgumentsValidated(Arguments&& cliArgs)
+    {
+        ApplyArguments(GetDefaultArguments());
+
+        // Read the config argument first and apply the config settings
+        if(cliArgs.configFilePath.is_initialized())
+        {
+            // Configuration via json config
+            ApplyArguments(std::move(ParseConfig(cliArgs.configFilePath.get())));
```
rtobar, line 149: No need for a `std::move` here

rtobar (approved):
Again, looks good to me.

Other than some minor comments, the main thing that confused me a bit was the new third "level" of arguments, does it make sense to introduce it? I see that `Arguments` is built from both the command line and the config file, and then each is applied separately to `ArgumentsValidated` via `ApplyArguments`, but I have the suspicion this could be simplified. I think I'd try to somehow merge `Arguments` and `ArgumentsValidated` maybe, but that's just a comment.

calgray:
I was hoping the third layer isn't but it would mean I'd have do something like convert the parsed json into command line strings (CLIArguments) and moving the type conversions into ArgumentsValidated which will result in a larger constructor or method.

The better approach is for CLI11 to support the typed `Arguments` struct via the use of lambdas. I suspect this could be possible via `Option->transform()` and maybe a templated helper but I haven't looked much into it.

calgray:
CLI11 also has built in support for INI config files and JSON via the use of `config_formatter` demoed here: https://github.com/CLIUtils/CLI11/blob/master/examples/json.cpp

The same problem of supporting enums and MVDirections described above applies but would be nice to use.

PR #68, opened 2020-11-10
Support trace macro

calgray:
I've noticed that when TRACE is defined the verbosity enum fails to compile. Here I've just favoured lower-case enum literals.


rtobar (approved):
Good point, I hadn't realized about it while reviewing earlier either. Maybe it also makes sense to rename the macro to something more explicit like `LEAP_DUMP_MATRICES` or similar to avoid future clashes.

calgray:
I know TRACE and DEBUG are standard macros in C# and I'm hesitant to use underscore prefixes in case of conflicts with system libraries. I'll consider changing it as the dumping is sometimes particularly slow.

PR #69, opened 2020-11-10
File Output Setting

calgray:
This adds support for specifying a file output rather than printing to the terminal either via CLI or config settings. Currently this only ovewrites the file at the specified path and reports an error early if the specified file cannot be written to.

I've also started adding more to the project documentation too make it easier to understand the project layout and terminology.

### DIFF
--- a/src/icrar/leap-accelerate-cli/Arguments.cc
+++ b/src/icrar/leap-accelerate-cli/Arguments.cc
@@ -40,7 +40,7 @@ namespace icrar
         args.source = InputType::FILENAME;
         args.filePath = boost::none;
         args.configFilePath = boost::none;
-        args.outFilePath = boost::none;
+        args.outputFilePath = boost::none;
 
         args.stations = boost::none;
         args.directions = boost::none;
@@ -55,7 +55,7 @@ namespace icrar
         : source(std::move(args.source))
         , filePath(std::move(args.filePath))
         , configFilePath(std::move(args.configFilePath))
-        , outFilePath(std::move(args.outFilePath))
+        , outputFilePath(std::move(args.outputFilePath))
         , mwaSupport(std::move(args.mwaSupport))
         , readAutocorrelations(std::move(args.readAutocorrelations))
     {
@@ -126,6 +126,24 @@ namespace icrar
             throw new std::invalid_argument("only stream in and file input are currently supported");
             break;
         }
+
+        if(m_outputFilePath.is_initialized())
+        {
+            //std::cout << "writing to " << m_outputFilePath.get() + ".json" << std::endl;
+            m_outputFileStream = std::make_unique<std::ofstream>(m_outputFilePath.get() + ".json");
+            if(!m_outputFileStream->is_open())
+            {
+                std::stringstream ss;
+                ss << "failed to open file " << m_outputFilePath.get() << ".json" << std::endl;
+                throw exception(ss.str(), __FILE__, __LINE__);
+            }
+
+            m_outputStream = m_outputFileStream.get();
+        }
+        else
+        {
+            m_outputStream = &std::cout;
+        }
     }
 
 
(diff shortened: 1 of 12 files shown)


src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -95,6 +95,13 @@ namespace cuda
 
         // Flooring to remove incomplete measurements
         int integrations = ms.GetNumRows() / ms.GetNumBaselines();
+        if(integrations == 0)
+        {
+            std::stringstream ss;
+            ss << "invalid number of rows, expected >" << ms.GetNumBaselines() << ", got " << ms.GetNumRows();
+            throw icrar::file_exception(ms.GetFilepath().get_value_or("unknown"), ss.str(), __FILE__, __LINE__);
+        }
```
calgray, line 103: Without this there was a "cannot iterate scalears" error on some measurement sets that caused confusion

src/icrar/leap-accelerate-cli/Arguments.h
```diff
@@ -114,6 +114,9 @@ namespace icrar
          */
         std::unique_ptr<MeasurementSet> m_measurementSet;
         std::istream* m_inputStream = nullptr; // Cached reference to the input stream
+        
+        std::unique_ptr<std::ofstream> m_outputFileStream;
+        std::ostream* m_outputStream = nullptr;
```
rtobar, line 119: `m_outputStream` doesn't need to be stored separately, it's already contained in the `unique_ptr`

src/icrar/leap-accelerate-cli/Arguments.cc
```diff
@@ -126,6 +126,23 @@ namespace icrar
...
+        {
+            m_outputFileStream = std::make_unique<std::ofstream>(m_outputFilePath.get() + ".json");
+            if(!m_outputFileStream->is_open())
+            {
+                std::stringstream ss;
+                ss << "failed to open file " << m_outputFilePath.get() << ".json" << std::endl;
+                throw exception(ss.str(), __FILE__, __LINE__);
+            }
+
+            m_outputStream = m_outputFileStream.get();
```
rtobar, line 140: Not needed

src/icrar/leap-accelerate-cli/Arguments.cc
```diff
@@ -196,6 +213,16 @@ namespace icrar
...
     }
 
+    boost::optional<std::string> ArgumentsValidated::GetOutputFilePath() const
+    {
+        return m_outputFilePath;
+    }
+
+    std::ostream& ArgumentsValidated::GetOutputStream()
+    {
+        return *m_outputStream;
```
rtobar, line 223: `return *m_outputFileStream.get()`

rtobar (commented):
Looks good, except that I think there's no need to store `m_outputStream` given that you are already storing `m_outputFileStream`, which is what my comments are about.

src/icrar/leap-accelerate-cli/Arguments.cc
calgray: I store m_outputStream seperately since in some cases m_outputStream is set to std::cout

rtobar: Ahhhh right, my bad, I didn't see that the first time. All good then :)

rtobar (approved):

src/icrar/leap-accelerate-cli/Arguments.cc
rtobar: Thinking about it again, an alternative would be to only store `m_outputFileStream`, then implement `GetOutputStream()` like:

```c++
std::ostream &ArgumentsValidated::GetOutputStream()
{
    if (m_outputFileStream)
    {
        return *m_outputFileStream.get();
    }
    return std::cerr;
}
```

calgray: Pretty much, the only difference between the two is m_outputStream is acting as a cache variable

PR #72, opened 2020-11-12
Simplify Calibration Result

calgray:
Calibration result currently uses a combination casacore::Matrix and std::vector that can be simplified and remove unnessary conversion.


src/icrar/leap-accelerate/model/casa/CalibrateResult.h
```diff
@@ -41,25 +41,25 @@ namespace casalib
...
+            , m_direction(direction)
+            , m_data(std::move(data))
         {
 
         }
 
         const casacore::MVDirection GetDirection() const { return m_direction; }
-        const boost::optional<std::vector<casacore::Vector<double>>>& GetData() const { return m_data; }
-        int GetIntegrationNumber() const { return m_integration_number; }
+        boost::optional<std::vector<casacore::Vector<double>>> GetData() const { return m_data; }
```
rtobar, line 61: Why not const ref? Otherwise there will be potentially unnecessary data copying happening.

src/icrar/leap-accelerate/algorithm/cuda/PhaseRotate.cu
```diff
@@ -201,7 +202,7 @@ namespace cuda
         Eigen::MatrixXd dIntColumn = dInt(Eigen::all, 0); // 1st pol only
         assert(dIntColumn.cols() == 1);
 
-        cal.push_back(ConvertMatrix(Eigen::MatrixXd((hostMetadata.GetAd() * dIntColumn) + cal1)));
+        auto cal = Eigen::MatrixXd((hostMetadata.GetAd() * dIntColumn) + cal1);
         output_calibrations.emplace_back(direction, cal);
```
rtobar, line 206: I hadn't realized before, but you can probably `std::move(cal)`  then emplacing it into `output_calibrations`. Alternatively you can do `Eigen::MatrixXd((hostMetadata.GetAd() * dIntColumn) + cal1)` directly within the `output_calibrations.emplace_back`, or even simply `(hostMetadata.GetAd() * dIntColumn) + cal1)` like you do in `cpu::PhaseRotate`

rtobar (commented):
I saw a couple of potential memory copies, but maybe I don't know well the semantics of the classes involved? I see the build in xenial is still failing too (but we have Travis builds back! At least temporarily) so that needs to be addressed too.

src/icrar/leap-accelerate/model/cpu/CalibrateResult.h
```diff
@@ -76,28 +76,38 @@ namespace cpu
...
+         */
         const MVDirection GetDirection() const { return m_direction; }
-        const std::vector<casacore::Matrix<double>>& GetData() const { return m_data; }
+
+        /**
+         * @brief Get the calibration Vector for the antenna array in the specified direction
+         * 
+         * @return const Eigen::MatrixXd 
+         */
+        const Eigen::MatrixXd GetData() const { return m_data; }
```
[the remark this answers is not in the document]
calgray: Sorry I was copying some recommendations from clang-tidy which suggested a lot of passing by value and std::copy. I'll run the linter over this file check that the pass by values make sense

calgray:
I'm not quite sure why both xenial and focal aren't building together. I also build locally with g++ 9.3.0 focal and -werror produces no warnings for memcpy

PR #79, opened 2020-11-20
Optional Xenial Args

calgray:
CLI11 optional types don't appear to work when building on xenial so I've added an extra step and tested via a virtual machine

### DIFF
--- a/src/icrar/leap-accelerate-cli/main.cc
+++ b/src/icrar/leap-accelerate-cli/main.cc
@@ -89,20 +89,23 @@ int main(int argc, char** argv)
 #else
     boost::optional<std::string> readAutocorrelations;
     app.add_option("-a,--autocorrelations", readAutocorrelations, "Set to true if measurement set rows store autocorrelations");
-    rawArgs.readAutocorrelations = readAutocorrelations.is_initialized() ? std::stoi(readAutocorrelations.get()) : (boost::optional<int>)boost::none;
 
     boost::optional<std::string> minimumBaselineThreshold;
     app.add_option("-m,--minimumBaselineThreshold", minimumBaselineThreshold, "Minimum baseline length in meters");
-    rawArgs.minimumBaselineThreshold = minimumBaselineThreshold.is_initialized() ? std::stod(minimumBaselineThreshold.get()) : (boost::optional<double>)boost::none;
 
     boost::optional<std::string> verbosity;
     app.add_option("-v,--verbosity", verbosity, "Verbosity (0=fatal, 1=error, 2=warn, 3=info, 4=debug, 5=trace), defaults to info");
-    rawArgs.verbosity = verbosity.is_initialized() ? boost::lexical_cast<bool>(verbosity.get()) : (boost::optional<bool>)boost::none;
 #endif
 
     try
     {
         app.parse(argc, argv);
+
+#if !__has_include(<optional>)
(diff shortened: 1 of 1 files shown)


src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -89,20 +89,23 @@ int main(int argc, char** argv)
...
 #endif
 
     try
     {
         app.parse(argc, argv);
+
+#if !__has_include(<optional>)
+        rawArgs.readAutocorrelations = readAutocorrelations.is_initialized() ? std::stoi(readAutocorrelations.get()) : (boost::optional<int>)boost::none;
+        rawArgs.minimumBaselineThreshold = minimumBaselineThreshold.is_initialized() ? std::stod(minimumBaselineThreshold.get()) : (boost::optional<double>)boost::none;
+        rawArgs.verbosity = verbosity.is_initialized() ? boost::lexical_cast<bool>(verbosity.get()) : (boost::optional<bool>)boost::none;
```
rtobar, line 107: It still appears to me the casts in this last line are to `int`, not to `bool`. There is a call to `std::stoi` above instead of a `boost::lexical_cast`, and you are already using `stod` too, so I'd make this one a `stoi` too to keep things aligned.

Also in these three cases the `rawArgs.*` members should be set to `boost::none` already when `rawArgs` gets constructed, so there's no need to set them to `none` again. So I'd go for something like (which also looks more like what you've done in the rest of the code, IIRC):

```c++
if (readAutocorrelations.is_initialized())
{
     rawArgs.readAutocorrelations = std::stoi(readAutocorrelations.get());
}
if (minimumBaselineThreshold.is_initialized())
{
    rawArgs.minimumBaselineThreshold = std::stod(minimumBaselineThreshold.get());
}
if (verbosity.is_initialized())
{
    rawArgs.verbosity = std::stoi(verbosity.get());
}
```

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -89,20 +89,23 @@ int main(int argc, char** argv)
...
     boost::optional<std::string> verbosity;
     app.add_option("-v,--verbosity", verbosity, "Verbosity (0=fatal, 1=error, 2=warn, 3=info, 4=debug, 5=trace), defaults to info");
-    rawArgs.verbosity = verbosity.is_initialized() ? boost::lexical_cast<bool>(verbosity.get()) : (boost::optional<bool>)boost::none;
 #endif
 
     try
     {
         app.parse(argc, argv);
+
+#if !__has_include(<optional>)
```
rtobar, line 104: I still think this is a very indirect of testing the capabilities of the CLIArgs library, plus it's C++17 so it might break some compilers in the future.

rtobar (approved):
The build works, so it suggests the changes are fine. I left two comments though, feel free to address them or not, I'll approve this now anyway.

src/icrar/leap-accelerate-cli/main.cc
calgray: I'm not perfectly sure what compiler feature causes the compile error as boost::optional<string> works as a expected. Testing with clang may help narrow it down but that might need to wait until migrating the CI builds

calgray:
I think the best approach from here might be to instead use boost command line arguments as conditional compilation for is very tedious. I'll merge this and put a up a task to test optional bool and int with that libary before attempting to spend time simplifying with CLI11

PR #80, opened 2020-11-20
Add Documentation

calgray:
Some parts of the codebase are missing documentation so I've added docs to a number of locations. From here it should be feasible to start adding documentation for all new functions in pull requests.

### DIFF
--- a/src/icrar/leap-accelerate/common/eigen_extensions.h
+++ b/src/icrar/leap-accelerate/common/eigen_extensions.h
@@ -105,11 +105,11 @@ namespace icrar
     }
 
     /**
-     * @brief 
+     * @brief Reads a file containing a binary hash at @p filename and outputs to @p hash
      * 
-     * @tparam T 
-     * @param filename 
-     * @param hash 
+     * @tparam T the hash type
+     * @param filename the hash file to read
+     * @param hash output parameter
      */
     template<typename T>
     void read_hash(const char* filename, T& hash)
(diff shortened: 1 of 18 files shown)


src/icrar/leap-accelerate/common/eigen_extensions.h
```diff
@@ -63,34 +63,54 @@ namespace icrar
...
     }
 
-
+    /**
+     * @brief 
+     * 
+     * @tparam T 
+     * @param filename 
+     * @param hash 
+     */
```
rtobar, line 113: This looks rather empty

rtobar (approved):
It's all doc changes, and it looks right to me (except an empty documentation I found, which I flagged).

I tried to generate the documentation locally and noticed a few things, which can be left as work for the future:

 * `INPUT` is empty, so doxygen tries to scan the full repository looking for source, which takes a long time (because of all the external code that is checked out). I'd suggest setting this to `src` for faster turnaround.
 * Several of the `WARN_*` options are set to `NO`, so it would seem there's nothing wrong. I tried turning on `WARN_IF_UNDOCUMENTED`, `WARN_IF_DOC_ERROR` and `WARN_NO_PARAMDOC` and that resulted in 488 warnings.
 * The `doxygen` cmake target seems incorrect: it runs under the `src` directory in the *binary* cmake directory, where there's very little to document (only the generated `git_revision.cc` file).
 * The `Doxyfile` has HTML output turned off, so to test this locally I had to at least turn it on.

calgray:
I know the doxyfile gets removed in this: https://github.com/ICRAR/leap-accelerate/pull/60/files Currently I don't plan on touching the html generation until I'm aware that the daliuge integration is correct to avoid conflicts.

calgray:
It looks like #60 was implemented elsewhere. I've created a task for this here: https://github.com/ICRAR/leap-accelerate/issues/84

PR #81, opened 2020-11-20
Remove casa implementation

calgray:
Currently there aren't any required casacore functions that have not been translated to eigen3 and as such there isn't need to continue supporting the casa calibration implementation. To allow for faster reading from casacore measurement sets there is the possiblilty to use matrix templates with Eigen::Map<>


rtobar:
I tried building locally and got this:

```
/home/rtobar/scm/git/leap-accelerate/src/icrar/leap-accelerate/model/cpu/CalibrateResult.h(124): error #20: identifier "m_data" is undefined
```

This wasn't caught by the CI builds because they don't build in debug mode (understandably, because otherwise tests run too slowly) and the identifier is defined inside an `assert()` call. It might be worth adding a new Travis job that builds in debug, but doesn't run the tests (an idea for the future though).

src/icrar/leap-accelerate/core/compute_implementation.cc
```diff
@@ -42,7 +42,8 @@ namespace icrar
     {
         if(value == "casa")
         {
-            out = ComputeImplementation::casa;
+            LOG(info) << "argument 'casa' deprecated, use 'cpu' instead";
```
rtobar, line 45: `LOG(warning)`

.travis/install.sh
```diff
@@ -44,5 +44,5 @@ CMAKE_OPTIONS="$CMAKE_OPTIONS -DGSL_ROOT_DIR=${GSL_ROOT_DIR}"
 CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_CXX_FLAGS_DEBUG=${CMAKE_CXX_FLAGS_DEBUG} -O1"
 CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_BUILD_TYPE=RelWithDebInfo"
 cmake .. ${CMAKE_OPTIONS} || fail "cmake failed"
-make all -j2 VERBOSE=1 || fail "make failed"
+make all -j VERBOSE=1 || fail "make failed"
```
rtobar, line 47: This is currently not failing, but it can be dangerous in some situations with too many files. `-j` means "compile as many files in parallel as possible" rather than "compile as many files are there are CPUs on this machine"; in other words `-j` is unconstrained, so if your project had more files you can run out of memory, and the OS will kill your compilation, leading to a build failure.

I'd *suggest* (but up to you) sticking with the old `-j2` because in Travis that's how many CPUs you are given (I think). You can alternatively count the number of CPUs (e.g., `grep processor /proc/cpuinfo  | wc -l`) and use that.

rtobar (approved):
Nice to see a lot of code go away :). Some minor comments only, looks great.

.travis/install.sh
calgray: Only removed as I found some builds that appeared to be timing out in a later branch when running linting after tests. I'll explore this change more in that PR.

PR #83, opened 2020-11-20
Abstract Calibration Method

calgray:
This change adds a factory method for a leap calibrator that allows hiding compile time implementation details from leap-accelerate-cli. This also allows for optional cuda compilation.

Only when writing tests does a programmer need to be aware of gpu compilation macros.


.travis/before_install_xenial.sh
```diff
@@ -48,16 +48,13 @@ export PATH=$PATH:$CUDA_HOME/bin
...
-sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 20
-
 sudo update-alternatives --install /usr/bin/cc cc /usr/bin/gcc 30
 sudo update-alternatives --set cc /usr/bin/gcc
 
+sudo update-alternatives --remove-all g++
+sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-5 10
+sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 20
 sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++ 30
 sudo update-alternatives --set c++ /usr/bin/g++
```
rtobar, line 60: I was expecting to see an extra piece of logic here to avoid installing CUDA at all when `CUDA_ENABLED=OFF`; otherwise you might be under the impression that CUDA-less builds are working while in reality they might not.

src/icrar/leap-accelerate-cli/main.cc
```diff
@@ -143,28 +145,12 @@ int main(int argc, char** argv)
...
-        {
-            cpu::CalibrateResult result = icrar::cuda::Calibrate(args.GetMeasurementSet(), args.GetDirections(), args.GetMinimumBaselineThreshold(), args.IsFileSystemCacheEnabled());
-            cpu::PrintResult(result ,args.GetOutputStream());
-            break;
-        }
-        }
-        return 0;
+        auto calibrator = LeapCalibratorFactory().Create(args.GetComputeImplementation());
+        cpu::CalibrateResult result = calibrator->Calibrate(args.GetMeasurementSet(), args.GetDirections(), args.GetMinimumBaselineThreshold(), args.IsFileSystemCacheEnabled());
+        cpu::PrintResult(result, args.GetOutputStream());
```
rtobar, line 153: Maybe something shorter like

````c++
auto results = ...;
PrintResult(result, ....);
```

src/icrar/leap-accelerate-cli/tests/E2EPerformanceTests.cc
```diff
@@ -98,7 +98,9 @@ namespace icrar
             }
             else if(impl == ComputeImplementation::cuda)
             {
+#ifdef CUDA_ENABLED
                 auto result = cuda::Calibrate(*ms, ToDirectionVector(directions), 0.0, false);
+#endif
```
rtobar, line 103: If you upgarde to gtest 1.10 apparently you can do this:

```c++
#else
    GTEST_SKIP_("Library built without CUDA support, skipping CUDA test");
#endif
```

The the rest of the `ifdef`s in this file are unnecessary, and you don't need to manually skip/disable tests.

Of course if you change to a runtime check then this would be slightly different, but the same idea.

src/icrar/leap-accelerate-cli/tests/E2EPerformanceTests.cc
```diff
@@ -98,7 +98,9 @@ namespace icrar
             }
             else if(impl == ComputeImplementation::cuda)
             {
+#ifdef CUDA_ENABLED
```
rtobar, line 101: Can use runtime function instead of compilation-time macro, as per the suggestion in the global review comment.

src/icrar/leap-accelerate/CMakeLists.txt
```diff
@@ -102,28 +118,33 @@ if(${CMAKE_VERSION} VERSION_GREATER "3.16.0")
 endif()
 
 
+if(CUDA_ENABLED)
 # Request that the target be built with -std=c++14
-# As this is a public compile feature anything that links to the target
+  # As this is a public compile feature anything that links to the target
```
rtobar, line 123: Misaligned

src/icrar/leap-accelerate/model/cpu/MetaData.h
```diff
@@ -22,7 +22,12 @@
 
 #pragma once
 
-#include "cuda_runtime.h"
+#if __has_include(<cuda_runtime.h>)
```
rtobar, line 25: `#ifdef CUDA_ENABLED` aligns better with the rest of the code, and doesn't depend on a C++17 feature.

src/icrar/leap-accelerate/tests/algorithm/PhaseRotateTests.cc
```diff
@@ -114,7 +114,9 @@ namespace icrar
             }
             else if(impl == ComputeImplementation::cuda)
             {
+#ifdef CUDA_ENABLED
```
rtobar, line 117: Same comment regarding rutime v/s compile time check, same for the rest of the file.

src/icrar/leap-accelerate/tests/math/vector_tests.cc
```diff
@@ -53,10 +53,12 @@ class vector_tests : public testing::Test
     {
         if(useCuda)
         {
+#ifdef CUDA_ENABLED
```
rtobar, line 56: Same comment regarding rutime v/s compile time check, same for the rest of the file.

src/icrar/leap-accelerate/tests/model/MetaDataTests.cc
```diff
@@ -188,6 +188,7 @@ namespace icrar
             EXPECT_DOUBLE_EQ(2.1537588131757608, meta.GetConstants().GetChannelWavelength(0));
         }
 
+#ifdef CUDA_ENABLED
```
rtobar, line 191: Same comment regarding rutime v/s compile time check, same for the rest of the file.

rtobar (changes requested):
I tried this build locally *with* CUDA and it still works. I saw there is a new Travis job that compiles without CUDA support, but the Travis infrastructure still installs the CUDA libs. That's the only thing I'd say needs to be addressed before accepting, otherwise you might think CUDA-less builds are working while in fact they aren't.

About the code: although it does the work, I found the introduction of the factory and interface classes unnecessarily complex as neither contains any state (they simply call the corresponding cpu/cuda `Calibrate` function. This is only a suggestion, but I would have maybe collapsed the factory and interface implementations into a simpler `icrar::Calibrate` function (or similar) that does the call forwarding directly, something like:

```c++
namespace icrar {
cpu::CalibrationResults Calibrate(ComputeImplementation impl, ....)
{
  if (impl == cpu)
    return cpu::Calibrate(...);
  else if (impl == cuda)
#ifdef CUDA_ENABLED
    return cuda::Calibrate(...);
#else
    throw invalid_arg();
#if
}
```

A final idea, again just a suggestion and not something that would block this PR: tests (and therefore any user code using the library) still needs to know if the library was compiled with CUDA support *at compile time*. I'd suggest changing this to a runtime check, which is much more flexible, and easily achievable by adding a new top-level function like:

```c++
namespace icrar {
bool HasCUDA()
{
#ifdef CUDA_ENABLED
    return true;
#else
    return false;
#endif
}
}
```

Then you can invoke that function from the tests (or from you executable, when printing version information, etc) to find out if the library was compiled with CUDA support, instead of relying on the macro definition.

src/icrar/leap-accelerate-cli/tests/E2EPerformanceTests.cc
calgray: the signature `cuda::Calibrate()` does not exist if the macro is disabled. If I took the guards away from the header it will then produce a link error unless I go through and stub all public cuda functions in the project.

src/icrar/leap-accelerate/tests/algorithm/PhaseRotateTests.cc
calgray: here the cuda vector functions do not exist if CUDA_ENABLED  is not defined.

.travis/before_install_xenial.sh
calgray: updated

src/icrar/leap-accelerate-cli/main.cc
calgray: this is mostly style preference, in some cases it is preferred to mark the type when it's not easy to infer from the r-value

calgray:
I've mostly started adapting to a calibrator class to handle compute device state should there be to do things like check for memory leaks, abstract the checking of available memory, call cudaDeviceReset in destructor or handle multiple gpus but I can restore this for a later time should there be cuda deployment problems

src/icrar/leap-accelerate-cli/tests/E2EPerformanceTests.cc
rtobar: True, hadn't spotted that. On the other hand now that you have `icrar::Calibrate` you could invoke that here instead of repeating the `if/else if` logic.

src/icrar/leap-accelerate/tests/algorithm/PhaseRotateTests.cc
rtobar: Same comment as above: you are right, I hadn't spotted that there was CUDA-specific code invoked within the `ifdef`, but on the other hand now that you have `icrar::Calibrate` you can probably reuse it here.

src/icrar/leap-accelerate/tests/math/vector_tests.cc
rtobar: Disregard my previous comment, this seem like a legitimate need for the macro now.

src/icrar/leap-accelerate/tests/model/MetaDataTests.cc
rtobar: Again, I was too quick here and misjudged, there is code that needs to be protected by the macro.

rtobar (approved):

rtobar:
Nice to see the CUDA-less build working! I might even switch to that locally now that it's supported. I left a couple of more minor comments, but this can be merged now.

src/icrar/leap-accelerate-cli/tests/E2EPerformanceTests.cc
calgray: Only concern with updating gtest at the moment is that I need nvcc to able to compile gtest.h. I'll create an issue outlining the steps check while doing this.

calgray: See https://github.com/ICRAR/leap-accelerate/issues/89

PR #85, opened 2020-11-26
Revert Channel Calculation

calgray:
Richard has informed me that the channel calculation we had previously was correct. I've reverted the offset changes.


rtobar (approved):
If Richard is happy, I'm happy!

PR #87, opened 2020-11-30
Cuda UVW rotation

calgray:
UVW rotation can be performed in cuda with minimal transfer overhead as UVW is already copied to gpu memory. To do this I've split metadata down into buffers to reduce memory copying and calculate rotated UVW in cuda.

10 directions on SKA:
Before: PhaseRotate in 3.01s
After: PhaseRotate in 2.91s

The slowest part of calibration is now computing calibrations using Ad1 and summing in the complex domain.
total time per direction on SKA: 0.3865s
uvw rotate (cuda): 0.036876s (9.5%)
rotate visibilities (cuda):  0.011451s (3%)
Ad1 and summing (cpu): 0.316177s (82%)


calgray:
Moved to https://gitlab.com/ska-telescope/icrar-leap-accelerate/-/merge_requests/4

PR #88, opened 2020-12-01
Produce less output when downloading files

rtobar:
The output generated by wget is by default too verbose for
non-interactive terminals (like in Travis), which generates outputs that
are difficult to read. This commit reduces the amount of output
generated by wget by adding the "-nv" argument. I also took the
opportunity to de-duplicate some of the code.

### DIFF
--- a/testdata/install.sh
+++ b/testdata/install.sh
@@ -21,13 +21,14 @@
 # MA 02111-1307  USA
 #
 
-[ ! -d mwa ] && mkdir mwa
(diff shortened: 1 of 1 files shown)


calgray (approved):
Changes look good! I've wondered whether it's possible for for CI to extract the errors from build log for quick reading but it looks like the download log is taking up a lot of the report even with VERBOSE=1. I'll consider taking out verbose too after this since I think I only added it to help debug link errors.

PR #93, opened 2020-12-08
Cuda Matrix PseudoInverse

calgray:
Matrix SVD is a slow operation that can instead be computed in parallel with cuda. Benchmarks soon to come.


calgray:
Moved to https://gitlab.com/ska-telescope/icrar-leap-accelerate/-/merge_requests/6

